PR #17, opened 2016-05-11
Check multiple locations to download spells

battlemidget:
This provides the ability to pull from multiple locations for spells

### DIFF
--- a/conjure/download.py
+++ b/conjure/download.py
@@ -1,0 +1,68 @@
+import os
+from subprocess import run, CalledProcessError
+
+
+class DownloadException(Exception):
+    pass
+
+
+def remote_exists(path):
+    """ Verifies remote url archive exists
+    """
+    try:
+        run('wget --spider -q {}'.format(path), shell=True,
+            check=True)
+    except CalledProcessError:
+        return False
+    return True
+
+
+def download(src, dst):
+    """ Download and extract archive
+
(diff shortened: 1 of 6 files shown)


debian/conjure-up.install
```diff
@@ -1 +1,2 @@
 share/*                        usr/share/conjure-up/.
+etc/*
```
mikemccracken: does it matter if the file doesn't end in a newline?

conjure/download.py
```diff
@@ -0,0 +1,68 @@
...
+         exist.
+    """
+    try:
+        run("wget -qO- {} | bsdtar -xf - -s'|[^/]*/||' -C {}".format(
+            src, dst), shell=True, check=True)
+    except CalledProcessError as e:
+        raise DownloadException("Unable to download {}: {}".format(src, e))
+
+
+def cycle_locations(path):
```
mikemccracken: This name was confusing when I read it where it was being called.
I see what it does here, but I think it'd read better if it was called like 'get_remote_url' or something that makes it clearer what it will return…

mikemccracken:
only minor comments. 👍  to merge with or without any changes. up to you

debian/conjure-up.install
battlemidget: nah this is fine

PR #18, opened 2016-05-11
tweaks

battlemidget:
make spell dir conjure/metadata.json
remove use of sudo for pollinate entropy server


mikemccracken:
looks good to me, still need to verify that pinging entropy directly gets what we want

PR #62, opened 2016-06-02
Patch charmstore to search var

battlemidget:
Rename charmstore -> charmstore-search

### DIFF
--- a/conjure/controllers/newcloud/gui.py
+++ b/conjure/controllers/newcloud/gui.py
@@ -173,7 +173,10 @@ def finish(credentials=None, back=False):
     utils.pollinate(app.session_id, 'CA')
 
     __do_bootstrap()
-    return controllers.use('variants').render()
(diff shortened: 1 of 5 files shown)


conjure/controllers/newcloud/gui.py
```diff
@@ -173,7 +173,10 @@ def finish(credentials=None, back=False):
     utils.pollinate(app.session_id, 'CA')
 
     __do_bootstrap()
-    return controllers.use('variants').render()
+    if app.fetcher != "charmstore-search":
+        return controllers.use('variants').render()
+    else:
+        return controllers.use('deploy').render()
```
mikemccracken: this new branch is the opposite of all the other ones - if not charmstore-search, it goes to variants and all the others go to deploy. is this intentional?

battlemidget: nice catch ill fix this after lunch

PR #67, opened 2016-06-02
Fixes #66

### LINKED ISSUE #66: cleanup service status view
The original service status view is being altered a bit to be more to the point about the applications being deployed. This view will also not be sticking around like it does in the 0.x series.

battlemidget:
Displays a simplified list of services and their units being
deployed. Along with a workload status for continued feedback.

This does look identical to the steps controller but that will be the next change to make that controller a 'post processing' controller/view


battlemidget:
image

PR #71, opened 2016-06-03
add steps logic, few drive by fixes

### LINKED ISSUE #29: post-deploy actions view MVP

battlemidget:
Big addition is the steps view and controller logic.

Also added support for waiting for at least half of the applications
to be in an active state before proceeding to the steps view.

Small cosmetic fixes to both deploystatus and steps views.

Fixes #29 

### DIFF
--- a/conjure/controllers/deploystatus/gui.py
+++ b/conjure/controllers/deploystatus/gui.py
@@ -5,11 +5,14 @@ from functools import partial
 from conjure import async
 from conjure.app_config import app
 from conjure import utils
+from conjure import controllers
+from conjure.api.models import model_info
 from . import common
 import os.path as path
 import os
 import json
 import sys
+import time
 
 
 this = sys.modules[__name__]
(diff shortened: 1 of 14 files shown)


conjure/controllers/deploystatus/gui.py
```diff
@@ -24,7 +27,13 @@
...
 
 def finish():
-    pass
+    try:
+        while not common.check_statuses():
+            app.log.debug('waiting for services to start')
+            time.sleep(10)
+        return controllers.use('steps').render()
+    except Exception as e:
+        return __handle_exception('ED', e)
```
battlemidget: will fix this to check a `done-deploy` script provided by the spells.

mikemccracken:
@battlemidget this steps view looks non-interactive. I thought the idea was that for each step, the user would read what it does and maybe see the juju command it was running (if shortcmd was set), and then hit enter - and only then is the script run.

Is that the eventual intention or were you thinking it should stay automatic like this?

battlemidget:
yea eventually the steps view will prompt for interactive input if required. Right now its just to get from start to finish for this milestone

PR #80, opened 2016-06-06
Make sure to switch to the most current model

battlemidget:
Always make sure to switch to the most current model
after bootstrap and in cases where a controller
exists and just new models are being added.


conjure/controllers/deploystatus/common.py
```diff
@@ -34,6 +34,7 @@ def wait_for_applications(script, error_cb, msg_cb):
                             result['message']))
                         count += 1
                     continue
+                count = 0
```
battlemidget: cosmetic to make sure the current agent status being checked is displayed to the user in the footer

PR #89, opened 2016-06-07
async handle deploystatus

battlemidget:
put wait_for_applications in the async queue and execute
callback once services are active

### DIFF
--- a/conjure/controllers/deploystatus/gui.py
+++ b/conjure/controllers/deploystatus/gui.py
@@ -3,6 +3,8 @@ from ubuntui.ev import EventLoop
 from conjure.app_config import app
 from conjure import utils
 from conjure import controllers
+from conjure import async
+from functools import partial
 from . import common
 import os.path as path
 import os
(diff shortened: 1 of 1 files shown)


conjure/controllers/deploystatus/gui.py
```diff
@@ -23,18 +25,30 @@ def __fatal(error):
     return __handle_exception('ED', Exception(error))
 
 
-def finish():
+def __handle_exception(tag, exc):
+    utils.pollinate(app.session_id, tag)
+    app.ui.show_exception_message(exc)
+
+
+def __wait_for_applications_exception(exc):
```
mikemccracken: fwiw, elsewhere we've been using partial(__handle_exception, "ED") instead of creating a whole wrapper. not a big deal though, just nice to be consistent

mikemccracken: (could do the same with __fatal too)

conjure/controllers/deploystatus/gui.py
```diff
@@ -23,18 +25,30 @@ def __fatal(error):
...
+                                  app.ui.set_footer),
+                          __wait_for_applications_exception)
+    future.add_done_callback(finish)
 
-def __handle_exception(tag, exc):
-    utils.pollinate(app.session_id, tag)
-    app.ui.show_exception_message(exc)
+
+def finish(future):
+    future.result()
```
mikemccracken: why call result() here? wait_for_applications doesn't actually return anything anyway, unless there's an error (and in that case it just returns the result of calling the callback, which isn't going to be anything useful anyway)

PR #108, opened 2016-06-09
Patch better step queue

battlemidget:
Properly waits for all steps to complete and returns a single result list to the summary view.

### DIFF
--- a/conjure/controllers/steps/common.py
+++ b/conjure/controllers/steps/common.py
@@ -1,5 +1,10 @@
 from subprocess import run, PIPE
 from conjure.app_config import app
+from conjure.api.models import model_info
+import time
+import json
+import os
+from collections import deque
 
 
 def __readlines_key(key, data):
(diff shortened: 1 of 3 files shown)


conjure/controllers/steps/common.py
```diff
@@ -49,3 +54,61 @@ def parse_title(step):
...
+    # Set our provider type environment var so that it is
+    # exposed in future processing tasks
+    app.env['JUJU_PROVIDERTYPE'] = info['ProviderType']
+
+    results = []
+    steps_queue = deque()
+    for step in steps:
+        if "00_pre.sh" in step \
+           or "00_post-bootstrap.sh" in step \
+           or "00_deploy-done.sh" in step:
```
mikemccracken: I've seen code for ignoring these non-step scripts in a few places now...
Is there a reason these scripts end up in steps? 
could we name them something else that doesn't get slurped up by the glob? maybe the glob could look for ("step-*.sh") and these could be named something else? step-NN-whatever.sh would still sort correctly, and we could simplify some of this code

PR #111, opened 2016-06-09
issue message callback for deploy/relation

### LINKED ISSUE #99: each deploy button press should indicate something is being deployed
The status bar should show some sort of message indicating services are still being deployed in the background before being able to show the deploy status

battlemidget:
Give user feedback on a single service deploy along
with relations being set. Note: this isn't necessary
on a headless mode install as we deploy the bundle
directly.

Fixes #99


conjure/juju.py
```diff
@@ -300,6 +307,8 @@ def do_add_all():
         for a, b in list(relations):
             params = {"Endpoints": [a, b]}
             try:
+                if msg_cb:
+                    msg_cb("Setting application relations")
```
mikemccracken: do we want this to say which relation? If not, let's just put it outside the 'for a,b in list(relations)' loop

mikemccracken:
@battlemidget do we want to show a message when the deploy / relation setting is complete, also? can use the same msg_cb and just say "Deploying {} ... done" or whatever

PR #123, opened 2016-06-10
Feature interactive steps

### LINKED ISSUE #91: steps view should be interactive
add a button to step through the steps so the user can read before it happens

### LINKED ISSUE #113: prefix steps filenames
Right now we glob for *.sh and skip other initialization scripts like 00_deploy-done.sh and 00_post-bootstrap.sh. It's done in at least 3-4 places now so lets prefix steps with `step-XX_name.sh` and rework the glob code to look for `step-*` and still have it ordered correctly.

battlemidget:
This allows spell authors to define additional inputs to be
presented to the user after deployment. For example, setting
a blog url for the Ghost charm, or telling Glance which Ubuntu
series to import images for.

Fixes #113
Fixes #91

image


conjure/controllers/steps/common.py
```diff
@@ -1,114 +1,57 @@
...
 
 
-def wait_for_steps(steps, message_cb, icon_state=None):
-    """ Waits for post processing steps and return its results
+def do_step(step, message_cb, icon_state=None):
+    """ Processes steps in the background
 
     Arguments:
-    steps: list of steps to run
+    step: list of steps to run
```
mikemccracken: need to update the 'list of steps' comment

conjure/ui/views/steps.py
```diff
@@ -1,60 +1,84 @@
...
+        cb: process step callback
+        """
         self.app = app
         self.table = Table()
+        self.cb = cb
 
         self.steps_queue = steps
         for k, v in self.steps_queue.items():
-            self.steps_queue[k] = StepWidget(v)
+            meta = StepWidget(v['step_metadata'])
```
mikemccracken: calling the StepWidget 'meta' is a pretty generic name, makes it hard to follow the rest of the code.

conjure/ui/views/steps.py
```diff
@@ -1,60 +1,84 @@
...
 
-    def update_step_message(self, key, msg):
-        """ Updates the return results from a step
-        """
-        self.steps_queue[key].result.set_text(('info_context', msg))
+    def done(self, *args):
+        self.cb({}, done=True)
+
+    def submit(self, btn, data):
+        self.cb(data)
```
mikemccracken: along the same lines as the comment above about 'meta', just calling this 'data' doesn't make it clear that it's a StepWidget object (at least I'm pretty sure it will be?)...

So when looking into what we're actually passing to finish() above, you need to look a lot deeper than if this had been `def submit(self, btn, stepwidget)`

mikemccracken:
StepWidget, and the different names used for StepWidgets when it's passed around confused me. 
StepWidget isn't a Widget subclass, it's a kind of model class with some view stuff (urwid Text objects) mixed in. At first I thought it was a widget because of its name and where in the tree it is, and I thought it's weird how it was being passed to the done callback and kind of used like a model object, ie in steps/gui.py finish(), 'data' is actually just a StepWidget AFAICT. 

I think StepWidget would be clearer as just a model class, called Step or StepInfo or something, and the title/description/result/icon urwid objects could be either factored out into a real StepWidget class, or just created on the fly in the loop in StepsView.

battlemidget:
k i can clean it up

battlemidget:
Here is an OpenStack interactive step list:

image

Just for reference here is the summary after the interactive steps:

image

PR #140, opened 2016-06-14
Rework step findings

battlemidget:
Parse steps that end with yaml and verify that a script is
found and is executable, log the error if not.

### DIFF
--- a/conjure/controllers/steps/common.py
+++ b/conjure/controllers/steps/common.py
@@ -3,6 +3,8 @@ from conjure.app_config import app
 from conjure.api.models import model_info
 import json
 import os
+from collections import deque
+from glob import glob
 
 
 def run_script(path):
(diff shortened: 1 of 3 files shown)


mikemccracken:
Looks good. As for running steps in tui mode, we do want that. maybe we should punt to the step metadata and define defaults there and just let tui mode use those defaults. Or we could have an environment variable that tells the script that it is running as tui. Then it could do whatever it needs to work.

battlemidget:
> Looks good. As for running steps in tui mode, we do want that. maybe we should punt to the step metadata and define defaults there and just let tui mode use those defaults. Or we could have an environment variable that tells the script that it is running as tui. Then it could do whatever it needs to work.

Ok, I'll create an issue to get the steps implemented in TUI mode

PR #147, opened 2016-06-14
Feature python hooklib

battlemidget:
Start working on a simple hook library that can be used by spell authors during the processing of steps or waiting for services to become available.

### DIFF
--- a/conjure/controllers/deploystatus/common.py
+++ b/conjure/controllers/deploystatus/common.py
@@ -1,5 +1,5 @@
-from subprocess import run, PIPE
 from conjure.app_config import app
+from conjure import utils
 from subprocess import CalledProcessError
 import json
 import os
@@ -20,7 +20,7 @@ def wait_for_applications(script, msg_cb):
             rerun = True
             count = 0
             while rerun:
-                sh = run_script(script)
+                sh = utils.run_script(script)
                 try:
                     result = json.loads(sh.stdout.decode('utf8'))
                 except json.decoder.JSONDecodeError as e:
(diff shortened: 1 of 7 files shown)


conjure/controllers/deploystatus/common.py
```diff
@@ -21,7 +21,12 @@ def wait_for_applications(script, msg_cb):
             count = 0
             while rerun:
                 sh = run_script(script)
-                result = json.loads(sh.stdout.decode('utf8'))
+                try:
+                    result = json.loads(sh.stdout.decode('utf8'))
+                except json.decoder.JSONDecodeError as e:
+                    app.log.exception(sh.stdout.decode())
+                    raise Exception(sh)
```
mikemccracken: what is 'sh' and what happens when you raise Exception(sh) ?

battlemidget: sh is the result of run_script which is just a subprocess run (sh meaning shell i guess), that exception will print the result object of subprocess run

battlemidget: @mikemccracken this also bubbles up to the handle_exception

mikemccracken: A: it's a subprocess.CompletedProcess and it's ok to pass to Exception, it just uses its repr or something

share/hooklib/juju.py
```diff
@@ -0,0 +1,24 @@
...
+
+def leader(application):
+    """ Grabs the leader of a set of application units
+
+    Arguments:
+    application: name of application to query.
+    """
+    sh = run('juju run --application $1 is-leader --format yaml',
+             shell=True, stdout=PIPE)
+    leader_yaml = yaml.load(sh.stdout.decode())
```
mikemccracken: can you add a check for success and log an error message if the run() fails? as-is this will send a crappy yaml traceback up the chain.

up to you if you want to then raise a new exception subclass like 'JujuException' or just return None or something

PR #152, opened 2016-06-14
Improve finding existing controller & adding model

### LINKED ISSUE #139: juju add-model call should use juju async queue
instead of blocking the UI when adding a model to an existing controller, we should just schedule it as an async job just like we do with bootstrapping a new controller

mikemccracken:
## PR comments:

This is two changes that went together. I wanted to background creating a new model.
However, when testing, I would call 'destroy-model' on the current model and then try again without manually switching, this ended up bootstrapping a new controller even though the controller was already there. I looked at the controller_provides_ctype function and couldn't tell why it was comparing model types instead of just controller cloud names, since in the clouds screen we are picking cloud names, not models. It also calls switch, which calls login() (I didn't understand why it did that either), and login() fails if there is no selected model, because it tries 'juju status' to check if juju is 'available'. Juju status only works if there's a selected model.

Hopefully this change is an improvement, but I've only had time to test it on 'localhost' (LXD) clouds, so maybe the other code was there to work around issues on other clouds? Let me know!

Fixes #139.

## commit log message:

Finds existing controller even if no current model or controller is
selected. Doesn't login(), so doesn't require 'juju status' to be
working to find a controller - it won't be if no model is selected.

Backgrounds creating new model on found controller, so UI is more
responsive. It's on the juju queue, so deploys will wait for add-model
to be done.

### DIFF
--- a/conjure/controllers/clouds/common.py
+++ b/conjure/controllers/clouds/common.py
@@ -51,9 +51,8 @@ def list_clouds():
     return sorted(list(clouds))
 
 
-def controller_provides_ctype(cloud):
-    """ Returns a controller that is bootstrapped and fulfills
-    the cloud type.
+def get_controller_in_cloud(cloud):
+    """ Returns a controller that is bootstrapped on the named cloud
 
     Arguments:
     cloud: cloud to check for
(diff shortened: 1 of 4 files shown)


battlemidget:
I'll fix the Travis ci failure

PR #183, opened 2016-06-28
Fix deploy via juju api

battlemidget:
They deprecated the adding charm to api server before deploy and now
require this step to be handled by the user of the api. yay. This fix
allows us to deploy services again.


battlemidget:
Going to go ahead and merge this as it is a small 1 line addition and it's blocking my other work

PR #188, opened 2016-06-28
Fixes #174

### LINKED ISSUE #174: post-bootstrap is not run when controller exists
The post-bootstrap script is only run in newcloud. if the clouds controller detects an existing controller, it will reuse it and not run the script.

This results in openstack-novalxd not working when reusing an existing controller, as its lxd profile is not set correctly.

Maybe we need to just change that script to be a pre-deploy script, instead of post-bootstrap, and run it in both cases.

battlemidget:
The reason why we removed the pre-deploy in the first place is because
we couldn't guarantee that a bootstrap was completed before attempting
to login to the api server.

Fortunately, we've recently added async queues so this renders this
issue moot and we can associate our pre deploy tasks with the same async
queue as juju bootstrap. This lets us reliably run the predeploy tasks
after a bootstrap is complete and we can successfully query required
parameters like the provider type.

### DIFF
--- a/conjure/controllers/deploy/gui.py
+++ b/conjure/controllers/deploy/gui.py
@@ -1,15 +1,16 @@
 from functools import partial
-
+import json
 import sys
-
+import os
 from ubuntui.ev import EventLoop
-
+from subprocess import check_output
 from conjure import controllers
 from conjure import juju
+from conjure import async
 from conjure.app_config import app
 from conjure.ui.views.service_walkthrough import ServiceWalkthroughView
 from conjure import utils
-
+from conjure.api.models import model_info
 from .common import get_bundleinfo, get_metadata_controller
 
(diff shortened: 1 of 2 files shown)


mikemccracken:
This looks like it will run the pre deploy script once per render. So once
per service if the user goes through every service or just once if they
skip them, or something in between.

I was assuming we'd only want to run it once overall- so maybe we just need
a flag to say we've already queued it that we check at the start of render.
What do you think?

-Mike

mikemccracken:
Ive only seen it do it once and ive gone through both ways. Ill re-run and
double check to make sure

mikemccracken:
nope, youre right i just noticed it, ill fix it in a minute

battlemidget:
Weird it looks like my emails show up as you :x:

PR #228, opened 2016-07-08
Add machines if specified in bundle

mikemccracken:
just deploying machines with placement specs in them will not create the machines, we need to add them first.

NOTE - the author of a spell should verify that the bundle is correct using 'juju deploy bundle.yaml' first, since there are some cases where an incorrect bundle could just hang forever with no error message when deployed piece by piece in this way.

Specifically, when an application has a placement spec with a machine ID number that won't exist, like "1" for a bundle with only one machine. The single machine will be named "0".


battlemidget:
@mikemccracken just double checking this didn't need an updated sync from bundle placer

PR #230, opened 2016-07-12
fix constraints for add machines

battlemidget:
constraints are no longer a string but of type:
https://godoc.org/github.com/juju/juju/constraints#Value

### DIFF
--- a/conjure/juju.py
+++ b/conjure/juju.py
@@ -330,10 +330,20 @@ def add_machines(machines, msg_cb=None, exc_cb=None):
     supported key
 
     """
+    def _prepare_constraints(constraints):
+        list_constraints = constraints.split(' ')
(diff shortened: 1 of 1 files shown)


conjure/juju.py
```diff
@@ -330,10 +330,20 @@ def add_machines(machines, msg_cb=None, exc_cb=None):
     supported key
 
     """
+    def _prepare_constraints(constraints):
+        list_constraints = constraints.split(' ')
+        new_constraints = {}
+        if len(list_constraints) > 0:
```
mikemccracken: minor readability comment: you don't need to check len > 0 here, since the `for c in list_constraints` loop will just do nothing in that case, so you can leave it off and make the code more readable

PR #236, opened 2016-07-13
Patch gh 205

### LINKED ISSUE #205: more feedback prior to displaying the service status screen
There is a long blank screen during the actual deploying of application X (seen in the footer). We should add some more indicators that things are working in the blank real estate

battlemidget:
Fixes #205

### DIFF
--- a/conjure/controllers/bootstrapwait/gui.py
+++ b/conjure/controllers/bootstrapwait/gui.py
@@ -2,33 +2,13 @@ from conjure.ui.views.bootstrapwait import BootstrapWaitView
 from conjure.app_config import app
 from conjure import controllers
 from ubuntui.ev import EventLoop
-from conjure import async
-from conjure import utils
-from functools import partial
 import sys
 
 this = sys.modules[__name__]
 
 
-def __handle_exception(tag, exc):
-    utils.pollinate(app.session_id, tag)
-    return app.ui.show_exception_message(exc)
-
-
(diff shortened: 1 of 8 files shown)


conjure/controllers/deploy/gui.py
```diff
@@ -110,7 +110,10 @@ def finish(single_service=None):
                                app.ui.set_footer,
                                partial(__handle_exception, "ED"))
 
-        return controllers.use('deploystatus').render(f)
+        if app.bootstrap.running:
```
mikemccracken: If the bootstrap has already completed by the time we get here, this will probably still be true, so we'll call bootstrapwait.render, which will flash the KITT bar for a second before the done_callback it adds is run, which will switch the view.

I think what you want here is to check `if running is True and running.done() is False`

conjure/controllers/bootstrapwait/gui.py
```diff
@@ -0,0 +1,29 @@
...
+this = sys.modules[__name__]
+
+
+def finish(*args):
+    return controllers.use('deploystatus').render()
+
+
+def __refresh(*args):
+    this.view.redraw_kitt()
+    EventLoop.set_alarm_in(1, __refresh)
```
mikemccracken: I think you need to store the alarm this returns and cancel it in finish(), because otherwise we'll just redraw kitt forever even after we've set the body to a different view

PR #249, opened 2016-07-14
load the summary screen with -s

battlemidget:


mikemccracken:
Is this what we want? what if someone kills it while things are deploying, before running any steps? Will this wait correctly in that case or just fail?

It'd be nice if there was a flag to just start where you left off, I think that's ideal.

battlemidget:
Yea i thought we had talked about not returning the user back to the steps view and just to the summary

PR #278, opened 2016-07-21
Fix automatically running steps in TUI

battlemidget:
There were some assumptions made on the model being passed into the
step, specifically it having .model and .widget attributes.

This fixes that as a step model in headless mode is not accessed in the
exact same way (no gui components)


conjure/controllers/steps/common.py
```diff
@@ -43,14 +43,20 @@ def do_step(step, message_cb, gui=False):
     Step title and results message
     """
 
-    step.clear_button()
+    if gui:
+        step_model = step.model
+        step_widget = step.widget
+    else:
+        step_model = step
+        step_widget = None
```
mikemccracken: This part is a little awkward, passing two different types in as `step` and then sorting it out inside depending on another argument. In tui, you pass in a `StepModel` as `step`, but in gui you pass in the `StepWidget` via the StepWidget's callback calling `finish()`.

There are a couple ways to avoid this, one would be to change `do_step` to just take the model and widget as separate arguments so you could pass in the StepModel and the StepWidget in the gui case or just the StepModel and None in the tui case. Then when you read this function you will always know what type the args are without needing to do this translation step.

conjure/controllers/summary/tui.py
```diff
@@ -1,15 +1,14 @@
 from . import common
 from conjure import utils
-import sys
 
 
 class SummaryController:
     def finish(self):
-        sys.exit(0)
+        return
```
mikemccracken: if all we're doing is returning here, we could just not have finish() and not call it in render...

battlemidget: good point, will fix

PR #291, opened 2016-07-26
make it snappy

### LINKED ISSUE #290: make it snap installable
conjure-up should be installable via snap

battlemidget:
Fixes #290


battlemidget:
Just FYI this is to be ready for when we can interface with

battlemidget:
It'll eventually need to be done, doesn't hurt to have it in place when we need it.

PR #300, opened 2016-07-29
more snappy friendliness

### LINKED ISSUE #282: Support logging to flatfile
Started in #279 

It would be nice to support skipping syslog logging all together. Specifically when conjure is being used in the context of automation. 

We can work around #279 by mapping in a syslog dev device: /dev/log into the container. This works primarily in boot2docker, and other linux host scenarios. Skipping the socket and going to flat file has benefits of not requiring the end user to have a syslog service running, and directional output for archival purposes (if required). 

This can be by flag to override any configuration stemming from packaging or env var?

I'm fine with either, and feel that it should be a tuneable option.

battlemidget:
With the removal of most basic necessities we have to rework how we
store logging, config files, pretty much everything.

and making this a snap enabled application is a hard requirement.

Fixes #282 

### DIFF
--- a/conjureup/app.py
+++ b/conjureup/app.py
@@ -23,7 +23,6 @@ import os
 import os.path as path
 import sys
 import uuid
-import yaml
 import re
 
 
@@ -158,6 +157,8 @@ def main():
     spell_dir = os.environ.get('XDG_CACHE_HOME', os.path.join(
         os.path.expanduser('~'),
         '.cache/conjure-up'))
+    if not os.path.isdir(spell_dir):
+        os.makedirs(spell_dir)
 
     app.fetcher = fetcher(opts.spell)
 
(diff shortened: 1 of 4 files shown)


conjureup/app.py
```diff
@@ -182,30 +183,19 @@ def main():
...
         app.config = {'metadata': None,
                       'spell-dir': spell_dir,
                       'spell': spell}
 
         # Set a general description of spell
         definition = None
-        if path.isfile('/usr/share/conjure-up/keyword-definitions.yaml'):
-            with open('/usr/share/conjure-up/keyword-definitions.yaml') as fp:
-                definitions = yaml.safe_load(fp.read())
-                definition = definitions.get(spell, None)
```
mikemccracken: Are we losing the keyword definition functionality here? 
we should be able to put that into a file in the snap somewhere right?

if adding it back in is too much for this branch, let's file an issue to track adding it back in later

conjureup/ui/views/bootstrapwait.py
```diff
@@ -38,7 +38,7 @@ def redraw_kitt(self):
                                             'bootstrap.err')
         out = check_output("tail -n 10 {}".format(bootstrap_stderrpath),
                            shell=True)
-        self.output.set_text(out)
+        self.output.set_text(out.stderr.decode())
```
mikemccracken: check_output returns the output, not a process object, so we don't need this edit

battlemidget: This was causing the snappy run to error out with a traceback, something about the encoding of the output was screwing it up.

conjureup/app.py
battlemidget: So this is probably going to change as we move away from storing our spells in the charmstore. Should probably talk about this next week.

conjureup/ui/views/bootstrapwait.py
mikemccracken: huh. I don't see that in my non-snappy run, I wonder what's different.
check_output does return bytes, so we should decode it here, but I guess it gets translated in urwid somewhere along the line, and that must be different for snappy runs?

Still, this change should cause a new traceback, since bytes objects do not have a stderr attribute.

battlemidget:
reworking this a bit

PR #308, opened 2016-08-01
do not rely on wget

battlemidget:


mikemccracken:
works for me with 'conjure-up -d cs:~spells/ghost'

PR #311, opened 2016-08-02
Add lots of test scaffolding

mikemccracken:
Ensure that we're at least calling render and finish on each controller
for gui and tui.

Add a few more involved tests here and there.

This should help us catch more simple load-time errors than the lint catches.

The other goal is that this should make it easier to add new tests when we find new bugs, instead of having to write these scaffolds being such a pain that we don't do it…


battlemidget:
awesome!

PR #317, opened 2016-08-02
re-organization with steps hooklib

battlemidget:
This is just a reorg for the hooklib stuff the spells can rely on.


battlemidget:
I want to just focus on the python hooklib stuff and not the bash one. Since spells that use bash are ones I've wrote I will be refactoring them to use the python hooklib stuff.

PR #344, opened 2016-08-10
Fix application deploy

battlemidget:
Read local bundle or bundle-location in order to parse out applications
that need to be deployed.


conjureup/ui/views/spellpicker.py
```diff
@@ -40,7 +40,7 @@ def _build_widget(self):
         total_items = [
             Padding.center_60(HR())
         ]
-        for spell in self.spells:
+        for spell in sorted(self.spells):
```
mikemccracken: why sort the spells here? 
As it was, they should just be in the order they were found in the yaml, which we can control. we might want to be able to pick an order other than alphabetical

battlemidget: ok i can take this out

PR #347, opened 2016-08-10
expose lxc,lxd,juju binaries in a snap

battlemidget:
You can now run /snap/bin/conjure-up.juju status to see your juju status
output.

Also update readme.

### DIFF
--- a/snapcraft/snapcraft.yaml
+++ b/snapcraft/snapcraft.yaml
@@ -15,6 +15,12 @@ apps:
     command: python3
   shell:
     command: bash
+  juju:
+    command: juju
+  lxd:
+    command: lxd
+  lxc:
+    command: lxc
 
 parts:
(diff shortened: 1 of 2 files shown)


snapcraft/snapcraft.yaml
```diff
@@ -15,6 +15,12 @@ apps:
     command: python3
   shell:
     command: bash
+  juju:
+    command: juju
+  lxd:
+    command: lxd
```
mikemccracken: Unless I'm missing something, I think we probably don't need to include 'lxd', since users shouldn't need to run the 'lxd' program.

I was wondering if it was necessary because part of the container creation process involves calling 'lxd forkstart' but since the binary is renamed to conjure-up.lxd anyway, having it unexported doesn't affect that

PR #364, opened 2016-08-23
clean up makefile for snappy

### LINKED ISSUE #321: clean up makefile for snappy builds
Need to remove the deb packaging targets and replace with snapcraft for building, pushing releases.

Basically the process is:

```
# Make sure to snapcraft login before doing this
snapcraft login
# enter creds
cd snapcraft
snapcraft build
snapcraft push conjure-up_2.0.0.7_amd64.snap --release stable
```

battlemidget:
Fixes #321


Makefile
```diff
@@ -5,7 +5,7 @@ NAME = conjure-up
 CURRENT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 TOPDIR := $(shell basename `pwd`)
 GIT_REV := $(shell git log --oneline -n1| cut -d" " -f1)
-VERSION := $(shell ./tools/version)
+VERSION := 2.0.0.8
```
mikemccracken: suggest we replace this with this:
```
$(shell python3 -c "import conjureup; print(conjureup.__version__)")
```

battlemidget: So the Makefile is the source for determining the version since it does some sed replacing lower in the `release` target

mikemccracken: oh ok. the change below had me confused about which one was the source.

PR #375, opened 2016-08-30
Fixes #374

### LINKED ISSUE #374: spells not sorted in spellpicker view

battlemidget:


mikemccracken:
This is fine for now, but weren't we going to change it so the spell index file was an array and just use the array order?

PR #394, opened 2016-09-07
add spell selector hover descriptions

battlemidget:
The hover part works now where it'll change the spell's description.

@mikemccracken I attempted to mess with the frame's focus position on pressing the tab key but I'm apparently missing something.

image

### DIFF
--- a/conjureup/ui/views/spellpicker.py
+++ b/conjureup/ui/views/spellpicker.py
@@ -71,7 +71,7 @@ class SpellPickerView(WidgetWrap):
 
     def _build_buttons(self):
         cancel = menu_btn(on_press=self.cancel,
-                          label="\nQUIT\n")
+                          label="\n  QUIT\n")
         buttons = [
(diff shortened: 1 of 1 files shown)


conjureup/ui/views/spellpicker.py
```diff
@@ -1,60 +1,100 @@
...
 
     def _swap_focus(self):
-        w_count = len(self._w.body.contents) - 1
-        if self._w.body.focus_position >= 2 and \
-           self._w.body.focus_position < w_count:
-            self._w.body.focus_position = w_count
+        self.current_focused_frame = None
+        if self.current_focused_frame is None:
+            self.current_focused_frame = self.frame.footer
+            self.frame.footer.focus_position = 0
```
battlemidget: @mikemccracken i was hoping just setting focus position to 0 in each of the frame's section would cause the first widget in that frame to be selected

PR #398, opened 2016-09-09
add journald logging

battlemidget:
Prior to setting the cachedir we relied on rsyslog for our logging, this
made it easier to define loggers outside of the running conjure-up
process. In addition to logging to a file this adds support for logging
directly to systemd journald. This makes it possible for us to keep all
logs relating to conjure-up and spells in a single location for parsing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: python
-sudo: true
+sudo: false
 python:
   - "3.5"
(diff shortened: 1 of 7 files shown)


mikemccracken:
I guess the systems import breaks the Travis build? Maybe we need to check
for that and not import? What a pain.

PR #417, opened 2016-09-20
Avoid showing config button for nothing

### LINKED ISSUE #415: applicationlist - dont show configure button if subordinate
if there are no whitelisted options for the application we should hide the configure button, otherwise it just shows the header and cancel/accept changes button:

image

mikemccracken:
If there are no whitelisted options for a service, don't show the button
that brings up the config screen.

Fixes #415.


mikemccracken:
since we need to show the config screen to edit num_units, this needs to be updated. we can hide configure only in the cases where there are no whitelisted options AND it's a subordinate charm.

PR #431, opened 2016-09-22
Fixes #215

### LINKED ISSUE #215: display required environment variables in headless mode
deploying a spell in headless mode may require additional environment variables to be set in order to make some of the post deploy steps work.

We should go through the spells additional-options in the steps yaml to determine if we need to check for pre-existing environment variables prior to deploying. If so error out with what variables were expected.

battlemidget:
-bin-bash_001

Displays environment variables used in any of the additional inputs from
post processing steps. Allowing you to set those environment variables
on the CLI for more customized headless installs.

### DIFF
--- a/conjureup/app.py
+++ b/conjureup/app.py
@@ -8,9 +8,12 @@ import sys
 import uuid
 
 import yaml
+from prettytable import PrettyTable
+import textwrap
+from termcolor import colored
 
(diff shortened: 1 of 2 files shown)


battlemidget:
Merging this as it's not part of the core functionality of conjure-up and informational only

PR #441, opened 2016-09-23
Fixes #295

### LINKED ISSUE #295: make model explicit when running juju-cli commands or API calls
utilize `juju -m [controller:]model` when executing juju cli commands. Currently we switch to the model we want and then run juju. This has the effect of killing any other existing conjure-up that may be running in seperate model.

As for the API we always switch to the known current model and login to the api, we need to login to the API without switching models.

battlemidget:
Gives the ability to run multiple conjure-up installs, allows the user
to switch controllers/models without interupting the current conjure-up
deployment, basically isolating each deployment.

### DIFF
--- a/conjureup/juju.py
+++ b/conjureup/juju.py
@@ -29,7 +29,7 @@ this.USER_TAG = None
 def requires_login(f):
     def _decorator(*args, **kwargs):
         if not this.IS_AUTHENTICATED:
-            login(True)
+            login(force=True)
         return f(*args, **kwargs)
     return wraps(f)(_decorator)
 
@@ -511,26 +511,6 @@ def get_accounts():
         env = yaml.load(c)
         return env['controllers']
     raise Exception("Unable to find accounts")
-
-
-def model_by_owner(controller, user):
-    """ List model associated with user
-
-    Arguments:
-    user: username to query
-    controller: controller to work in
-
-    Returns:
-    Dictionary containing model information for user
-    """
-    models = get_models(controller)
-    for m in models:
-        if m['owner'] == user:
-            return m
-    raise LookupError(
-        "Unable to find user: {}".format(
-            user
-        ))
 
 
 def get_model(controller, name):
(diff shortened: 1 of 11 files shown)


conjureup/juju.py
```diff
@@ -29,7 +29,7 @@
 def requires_login(f):
     def _decorator(*args, **kwargs):
         if not this.IS_AUTHENTICATED:
-            login()
+            login(True)
```
mikemccracken: It's more readable to put the keyword in here, like `login(force=True)` , then you know what the True is for

conjureup/juju.py
```diff
@@ -559,16 +513,17 @@ def get_accounts():
     raise Exception("Unable to find accounts")
 
 
-def model_by_owner(user):
+def model_by_owner(controller, user):
```
mikemccracken: it looks like model_by_owner isn't used anywhere, and it's a little odd - it'll just return the arbitrary first model owned by 'user'.

What do you think about just getting rid of it?

test/test_controllers_clouds_gui.py
```diff
@@ -91,7 +91,9 @@ def test_finish_w_controller(self):
         self.mock_gcc.return_value = 'testcontroller'
         self.controller.finish('testcloud')
         self.mock_juju.assert_has_calls([
-            call.switch_controller('testcontroller')])
+            ANY,
+            ANY,
```
mikemccracken: What are the first two calls here? if possible we should test them, or else just avoid asserting anything about them, and use assert_called_once_with or something

conjureup/juju.py
battlemidget: Fixed

conjureup/juju.py
battlemidget: removed i did a quick grep and verified it wasn't ued anywhere

test/test_controllers_clouds_gui.py
battlemidget: fixed the test as well

PR #446, opened 2016-09-27
add pip3 to debian/control

battlemidget:
This is so spells like openstack-novalxd can install the openstack
clients python dependencies in the user's .local/lib directory. Bypasses
the needs for sudo apt install python3-openstackclient.


mikemccracken:
Is there a way we could use a virtualenv to avoid installing things into the user's home dir?
If someone was a python dev (but wasn't using venv themselves) and we installed new versions of libraries they were building with, that'd be frustrating.

battlemidget:
Yea this is what I did here:

https://github.com/conjure-up/spells/blob/master/openstack-novalxd/steps/step-03_keypair#L38

I think we can leave this change out now

battlemidget:
Oh this is still necessary to make sure pip is installed, we can make sure virtualenv gets installed as well though.

PR #457, opened 2016-09-30
Fixes #251

### LINKED ISSUE #251: do not allow button press on summary until all steps are complete
We don't want the summary to show up prematurely as it won't display all step results if pressed before the queue was empty.

battlemidget:

### DIFF
--- a/conjureup/controllers/steps/gui.py
+++ b/conjureup/controllers/steps/gui.py
@@ -13,8 +13,6 @@ from conjureup.ui.views.steps import StepsView
 from conjureup.ui.widgets.step import StepWidget
 from ubuntui.ev import EventLoop
 
-ASYNC_STEPS_QUEUE = 'async-steps-queue'
-
 
 class StepsController:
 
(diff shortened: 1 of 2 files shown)


battlemidget:
you're right it doesn't show the last output in the status. need to fix that

conjureup/controllers/steps/gui.py
```diff
@@ -51,6 +51,14 @@ def get_result(self, future):
         app.log.debug("Storing step result for: {}={}".format(
             step_model.title, step_model.result))
         self.results[step_model.title] = step_model.result
+        if len(self.view.steps) == 0:
+            app.log.debug(
+                "End of step list waiting for last step to complete "
+                "then rendering summary.")
+
+            async.submit(partial(self.finish, None, None, done=True),
```
battlemidget: If no steps remain and the last step returns quickly (less than a second) it will seem like it immediately switches to the status screen. We can either put a second delay here or in the spells that may return quickly.

PR #465, opened 2016-10-03
fix new-cloud path

mikemccracken:
Simplify clouds gui controller:

We no longer get here if there's an existing controller, so we don't
need to check for that.

We now only get here if the user chose 'create new' in the
controllerpicker.


mikemccracken:
thanks @battlemidget

PR #467, opened 2016-10-03
Fix 462

### LINKED ISSUE #462: need a way to specify an existing controller on CLI
eg. `conjure-up foo google existingcontroller`

mikemccracken:
Two changes:

1. add an arg to specify a controller within a cloud - Fixes #462.
2. check and show errors from bootstrap on TUI path

note that if you try to 'conjure-up spellname maas', it will complain that it doesn't know about maas. This is because we don't have a good way of passing the server name into maas. You need to bootstrap first, or just use the GUI to bootstrap onto a maas.


mikemccracken:
fixing tests rn

PR #468, opened 2016-10-04
Fixes #464

### LINKED ISSUE #464: blacklist not working for conjure-up canonical-kubernetes

battlemidget:
This addresses the issue where the blacklist wasn't being properly
rendered in the UI. It also will hide choosing a public cloud if maas is
the only thing in the whitelist metadata. This should cover all
scenarios where either a whitelist or blacklist is defined and if
neither are found in the metadata.


conjureup/controllers/clouds/gui.py
```diff
@@ -28,7 +32,9 @@ def render(self):
             "Please select from a list of available clouds")
         view = CloudView(app,
                          clouds,
-                         self.finish)
+                         blacklist=parse_blacklist,
+                         whitelist=parse_whitelist,
```
battlemidget: Done

PR #472, opened 2016-10-04
Fixes #124

### LINKED ISSUE #124: steps need basic interactive input validation
During the interactive steps we need to make sure that required inputs are validated so that the shell scripts can always rely on those environemnt variables

battlemidget:

### DIFF
--- a/conjureup/ui/widgets/step.py
+++ b/conjureup/ui/widgets/step.py
@@ -37,6 +37,7 @@ class StepWidget(WidgetWrap):
         self.result = Text(step_model.result)
         self.output = Text(('info_minor', ''))
         self.icon = Text(("info_minor", "\N{BALLOT BOX}"))
+        self.has_error = False
 
         self.additional_input = []
         if len(step_model.additional_input) > 0:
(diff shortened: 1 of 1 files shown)


battlemidget:
screenshot_2016-10-04_18-16-35

conjureup/ui/widgets/step.py
```diff
@@ -175,6 +176,19 @@ def generate_additional_input(self):
         self.step_pile.focus_position = self.current_button_index
 
     def submit(self, btn):
+        for i in self.additional_input:
+            if i['input'] and (
+                    i['input'].value is None and self.model.required):
+                self.app.log.debug("Missing default: {}".format(i))
```
mikemccracken: this log message is a little confusing, why 'missing default'?
maybe 'missing required input:' ?

conjureup/ui/widgets/step.py
```diff
@@ -175,6 +176,19 @@ def generate_additional_input(self):
...
+                self.has_error = True
+                self.step_pile.contents.insert(
+                    1,
+                    (Text(('error_major', 'Missing required input.')),
```
mikemccracken: Since this gets put in the first position, It'd be cool if the message said which input was missing. I guess just use the i['label'] ?

Now that I think of it, what happens if more than one input is missing? It looks like we'll just show a message about the first one. It'd be great if instead we could collect a list of labels that had missing inputs and then have the message say "Missing required input: foo, bar, baz".

battlemidget:
screenshot_2016-10-05_09-00-26

PR #481, opened 2016-10-06
prep 2.0.1 release

battlemidget:

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,12 +1,26 @@
-conjure-up (2.0.1~beta2) UNRELEASED; urgency=medium
+conjure-up (2.0.1) yakkety-proposed; urgency=medium
 
-  * 2.0 Development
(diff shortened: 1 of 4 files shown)


debian/changelog
```diff
@@ -1,12 +1,26 @@
-conjure-up (2.0.1~beta2) UNRELEASED; urgency=medium
+conjure-up (2.0.1) yakkety-proposed; urgency=medium
 
-  * 2.0 Development
+  * New upstream release
+  * autopkgtest support
+  * Full Juju 2.0 support
+  * Full MAAS 2.0 support
+  * Native Juju bundle support
```
mikemccracken: not sure how important this is, but it's not really clear what this means. Do you mean that now we support all the known keys in bundles? because we don't use the 'native' juju bundle deploying stuff in 2.0

PR #482, opened 2016-10-07
Add some extra options to grapher

mikemccracken:
These are really old changes that I never pushed up apparently.
I used them for the graph in the wireframe doc.


mikemccracken:
@battlemidget fixed the lint on here, mind taking another look?

PR #488, opened 2016-10-11
Add FLOAT type to option widget handling

mikemccracken:
Minimal validation, will not pass thru any invalid floats.


mikemccracken:
https://bugs.launchpad.net/conjure-up/+bug/1631770

PR #496, opened 2016-10-19
Fix maas apiclient to work with MAASv2 api

battlemidget:
Also embedded the api client in the code rather than syncing with
maasclient. This is a simple class to wrap get/post/put/delete requests
against the maas server. The querying is left up to the developer, for
example,

```
from conjureup.maas import MaasClient
m = MaasClient(server_address='127.0.0.1', consumer_key='...')
res = m.get('/nodes/')
res.ok
```


conjureup/controllers/newcloud/gui.py
```diff
@@ -132,6 +133,14 @@ def finish(self, credentials=None, back=False):
         if self.cloud == 'maas':
             self.cloud = '{}/{}'.format(self.cloud,
                                         credentials['@maas-server'].value)
+            # Store the MAAS information for the api client
+            maas_api_key = common.parse_maas_apikey(
+                credentials['maas-oauth'].value)
+            app.maas.client = MaasClient(
```
battlemidget: This is so you can pass the maas client through to bundleplacer

mikemccracken: OK, nice. I'll actually need it over in deploy/gui instead, since we don't always run the newcloud controller, but I'll do that in a separate branch

PR #503, opened 2016-11-10
Use google analytics

mikemccracken:
Add telemetry for improving user experience

Important info about how we're tracking responsibly:
* Allows turning it off with a CLI flag
* anonymizes IPs (sends the aip=1 param)
* sends which screens are seen, which spells are chosen, and errors. no info about cluster or configs.


battlemidget:
Very cool

PR #527, opened 2016-12-19
Improved ux for architecture view

mikemccracken:
1. makes pin button (juju machine ID) more obviously editable
2. makes constraint field more obviously editable
3. Adds help text about pin button
4. makes 'add new machine' button more visible by making it the same size as other buttons


mikemccracken:
I don't think this looks great on every terminal setting ,would like some feedback on the background color - I just matched the filter field.

image

PR #530, opened 2016-12-19
Fix 521

battlemidget:


conjureup/ui/views/lxdsetup.py
```diff
@@ -106,20 +106,17 @@ def buttons(self):
 
     def build_info(self):
         items = [
-            Text("There was no LXD bridge found on your system "
-                 "which usually means this is your first time running "
-                 "LXD."),
+            Text("There was no LXD bridge found on your system or "
+                 "you are not part of the LXD user group which usually "
```
mikemccracken: I think this would be better if we were exact, and said "not part of the unix group 'lxd', which ..."

conjureup/utils.py
```diff
@@ -136,6 +136,15 @@ def check_bridge_exists():
     return False
 
 
+def check_user_in_group(group):
+    """ Checks if a user is associated with `group`
+    """
+    groups = run_script('id').stdout.decode()
```
mikemccracken: I think we should use `groups` here instead of `id`. `id` returns more than just the groups and could break if a user was named 'lxd' but not in the 'lxd' group or some other kind of screwup

battlemidget: ah i like that better, +1 for sysadmin skillz

PR #542, opened 2017-01-06
We dont use this any longer (craft.py)

battlemidget:


mikemccracken:
thanks

PR #548, opened 2017-01-09
Registry changes

mikemccracken:
Two related changes made while figuring out how to solve https://github.com/conjure-up/spells/issues/25

1. some cleanup in the spells registry sync code

2. Keep the category for spells around, sort by that and show it in the list

Here's what it looks like:

image

### DIFF
--- a/conjureup/app.py
+++ b/conjureup/app.py
@@ -4,6 +4,7 @@
 import argparse
 import os
 import os.path as path
+import subprocess
 import sys
 import textwrap
 import uuid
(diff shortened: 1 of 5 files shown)


mikemccracken:
I definitely think we should sort using the category but I'm open to suggestions about leaving off the category names in the UI. They're lower case because they correspond to what you type on the command line, e.g. 'conjure-up openstack', which will show you the two openstack options...

battlemidget:
I really like this, maybe in an iteration call 'other' something else? Like maybe 'Additional curated' or something to that effect

PR #564, opened 2017-01-12
Ensure that bundle placement directives are always used and shown in UI

### LINKED ISSUE #553: Conjure-up uses double the deployed hardware when deploying openstack-base and kubernetes-core
cs:kubernetes-core uses 2 physical machines to deploy the services for kubernetes. Conjure-up bootstraps one node for juju, then 4 more for the services. Not sure why this is the case

mikemccracken:
Fixes #553 

Since #514, bundles that came with placement directives were only partially shown in the ui and the placement directives were ignored when deploying.
The bundle's machines and their constraints were used but the directives in the individual applications were ignored.

Test this by deploying openstack-base or kubernetes-core onto MAAS and note that without this branch they use one machine per service. With the branch they use 4 and 2 machines respectively, and many services are on lxds.

### DIFF
--- a/conjureup/controllers/deploy/gui.py
+++ b/conjureup/controllers/deploy/gui.py
@@ -122,7 +122,6 @@ class DeployController:
 
     def sync_assignments(self):
         bundle = app.metadata_controller.bundle
-        midx = 0
         for bundle_application in bundle.services:
             deployargs = bundle_application.as_deployargs()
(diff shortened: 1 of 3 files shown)


mikemccracken:
@canonical-cgregan this PR should fix the issues you noticed. Thanks for the bug report!

PR #581, opened 2017-01-14
Snappy updates

battlemidget:
- improve network and sysctl settings for localhost.
  Sets up a network bridge via systemd along with for
  the current environment so users can deploy to localhost
  immediately.
- pull in lxc/lxd as parts dependencies.
  We utilize the LXD snap part so it is built in the
  same exact way as upstream. This will allow us to
  stabilize conjure-up on a known set of dependency versions
  even across distro series.
- correct proper paths in a snap environment.
  in classic mode this allows us to utilize the filesystem
  as normal and makes the transition to snap much easier.

### DIFF
--- a/conjureup/log.py
+++ b/conjureup/log.py
@@ -27,24 +27,20 @@ class _log:
 
 
 def setup_logging(app, logfile, debug=False):
-    try:
-        cmdslog = TimedRotatingFileHandler(logfile,
-                                           when='D',
-                                           interval=1,
-                                           backupCount=7)
-    except FileNotFoundError as e:
-        # SNAP: This failing
-        # May need to do something different for snappy
-        cmdslog = logging.FileHandler('deploy-done.log')
+    cmdslog = TimedRotatingFileHandler(logfile,
+                                       when='D',
+                                       interval=1,
+                                       backupCount=7)
 
     if debug:
         env = logging.DEBUG
         cmdslog.setFormatter(logging.Formatter(
-            "%(name)s: [%(levelname)s] %(filename)s:%(lineno)d - %(message)s"))
+            "%(asctime)s %(name)s: [%(levelname)s] "
+            "%(filename)s:%(lineno)d - %(message)s"))
     else:
         env = logging.INFO
         cmdslog.setFormatter(logging.Formatter(
-            "%(name)s: [%(levelname)s] %(message)s"))
+            "%(asctime)s %(name)s: [%(levelname)s] %(message)s"))
 
     cmdslog.setLevel(env)
 
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -1,8 +1,52 @@
 #!/bin/bash
 
+
+VERSION=$(lxd --version)
+
+# LXD 2.3+ needs lxdbr0 setup via lxc.
+if dpkg --compare-versions "$VERSION" gt "2.2"; then
+    if ! lxc network list | grep -q lxdbr0; then
+        # Configure a known address ranges for lxdbr0.
+        lxc network create lxdbr0 \
+            ipv4.address=10.0.8.1/24 ipv4.nat=true \
+            ipv6.address=none ipv6.nat=false
+    fi
+    lxc network show lxdbr0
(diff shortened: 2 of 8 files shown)


battlemidget:
I've tagged the last pre-snappy branch so we can push forward with getting conjure-up ready for a snap install.

conjureup/log.py
```diff
@@ -27,24 +27,20 @@ def warning(self, msg):
...
-                                           interval=1,
-                                           backupCount=7)
-    except FileNotFoundError as e:
-        # SNAP: This failing
-        # May need to do something different for snappy
-        cmdslog = logging.FileHandler('deploy-done.log')
+    cmdslog = TimedRotatingFileHandler(logfile,
+                                       when='D',
+                                       interval=1,
+                                       backupCount=7)
```
mikemccracken: @battlemidget i'm curious what changed that this works now

conjureup/log.py
```diff
@@ -27,24 +27,20 @@ def warning(self, msg):
...
         env = logging.DEBUG
         cmdslog.setFormatter(logging.Formatter(
-            "%(name)s: [%(levelname)s] %(filename)s:%(lineno)d - %(message)s"))
+            "%(asctime)s %(name)s: [%(levelname)s] "
+            "%(filename)s:%(lineno)d - %(message)s"))
     else:
         env = logging.INFO
         cmdslog.setFormatter(logging.Formatter(
-            "%(name)s: [%(levelname)s] %(message)s"))
+            "%(asctime)s %(name)s: [%(levelname)s] %(message)s"))
```
mikemccracken: 👍

snap/hooks/configure
```diff
@@ -1,29 +1,61 @@
...
+set lxd/bridge-ipv4-nat true
+set lxd/bridge-ipv6 false
+EOF
+
+    rm -rf /etc/default/lxd-bridge
+
+    dpkg-reconfigure lxd --frontend=noninteractive
+
+    # Must run a command for systemd socket activation to start the service
+    lxc config show
```
mikemccracken: I'm surprised we need to do this. IIRC this was required a long time ago to make sure that lxd's bridge was set up or something, but lxd has changed quite a bit about that. Are you sure we still need this?

snapcraft/script/bridge.stop
```diff
@@ -3,6 +3,8 @@
 ip addr flush dev conjureup0
 ip link set dev conjureup0 down
 
+sleep 5
+
```
mikemccracken: seems error prone, is there any thing we can loop on to be sure we've waited long enough?

snapcraft/script/bridge.start
```diff
@@ -1,11 +1,15 @@
 #!/bin/sh
 
+logger -t "conjure-up/systemd" "[DEBUG] Creating conjure-up network bridge"
 ip link add dev conjureup0 type bridge
 ip addr add 10.99.0.1/24 dev conjureup0
 ip link set dev conjureup0 up
 
+sleep 5
+
```
mikemccracken: seems error prone, is there any thing we can loop on to be sure we've waited long enough?

snap/hooks/configure
battlemidget: I tested with and without setting up a bridge and thia worked everytime

conjureup/log.py
battlemidget: The snappy wrapper was overriding $HOME once I fixed that this worked like before

snapcraft/script/bridge.stop
battlemidget: Yea need to clean this up, snappy and systemd have been giving me issues

snap/hooks/configure
mikemccracken: sorry, I wasn't clear - I just meant line 37, the 'lxc config show', to start the service. why does the service need to be started? I think this might be trouble, since IIRC we can't be sure that it'll stay up for any particular length of time after we do this poke. If we need it to be up for something else, we need to do something there instead, maybe.

PR #588, opened 2017-01-17
Fixes #583

### LINKED ISSUE #583: It's not clear whether I should use the mouse or keyboard
Apparently both is possible but there is little indication of what happens when. 
Sometimes the installer appears to be stuck but when i click inside, suddenly it appears i have activated some hidden option and it continues.

battlemidget:
We don't heavily test the use of a mouse within the conjure-up terminal UI.
Disable this feature to limit any unexpected results when a mouse click is
registered inside the app.

### DIFF
--- a/conjureup/app.py
+++ b/conjureup/app.py
@@ -319,6 +319,7 @@ def main():
     else:
         app.ui = ConjureUI()
         EventLoop.build_loop(app.ui, STYLES,
-                             unhandled_input=unhandled_input)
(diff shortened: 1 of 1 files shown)


mikemccracken:
@battlemidget let's discuss this - I'm not sure it's necessary, and we don't have a good grasp on how many people use the mouse. IMO some of the screens are better with the mouse, even if I don't often test with it.

PR #618, opened 2017-01-19
Fix 541

mikemccracken:
The main change is that ensure_machines is now async, which means that it can block on adding a machine without blocking the UI.

This also adds more feedback to the status line, so you'll see when machines are being added there, to help explain why the deploystatus screen might be empty for a while.

### DIFF
--- a/conjureup/async.py
+++ b/conjureup/async.py
@@ -5,7 +5,7 @@ work.
 
 import logging
 import time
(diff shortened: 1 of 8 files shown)


battlemidget:
I'm out today and tom but wanted to get this merged, I tested the branch and didn't run into anything and the deploy blocking the UI is fixed

PR #619, opened 2017-01-23
Fixes #595

### LINKED ISSUE #595: conjure-up uses petname for model name
conjure-up apparently uses single-word petnames for Juju models it
creates, e.g.:
```
| james@malefic:~$ juju models
| Controller: lxd
|
| Model Cloud/Region Status Machines Cores Access Last connection
| controller localhost/localhost available 1 - admin just now
| default localhost/localhost available 0 - admin 2017-01-02
| dane* localhost/localhost available 4 - admin 6 minutes ago
|
| james@malefic:~$
```
A bare pet name is pretty confusing and would only get more so if you
conjured up multiple spells. At a minimum, I think the spell should
be indicated in some way in the model name (perhaps combined with a
petname, if you need to avoid clashes). It might also be useful to
indicate in some way that this is a conjure-up model?

battlemidget:
Just need the review right now, going to test and make sure it does what we want.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,15 @@ $ sudo apt update
 $ sudo apt install conjure-up
 ```
 
+## alternate (soon to be recommended)
+
+We will be switching over to a snap based installation in the very near future.
+Please help us test this out by installing via `snap` command:
+
+```
+$ sudo snap install conjure-up --classic --edge
+```
+
 # how to use
 
 ## Run the installer interactively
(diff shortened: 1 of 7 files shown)


mikemccracken:
A couple minor comments:

I think maybe the model name should include 'conjure-up' also, so when you conjure-up and use a preexisting controller, you still know that the model came from conjure-up.

It might also be nice to include cloud type in the controller name, like 'conjure-up-aws'. 

Also I don't think we need more than 3-4 digits of random UUID string. You get 8 from the code you wrote, maybe just truncate it? Someone's going to end up typing the whole thing and get mad..

battlemidget:
@mikemccracken updated based on your suggestions

battlemidget:
@mikemccracken should be ready for review and merge, test deploys reflects both the updated controller and models

PR #625, opened 2017-01-26
Add conjure-down

### LINKED ISSUE #65: make cancel/quit button show a dialog box giving option to kill model
When quitting out of conjure-up on any view besides the summary view and any error/exception view to give the user the option to remove the model.

### LINKED ISSUE #587: If lxc is used, conjure-up needs to deal with lxc problems too
If there are lxc containers left over from interrupted installations, conjure-up should allow me to view and delete them. 
If lxc networking isnt setup, conjure-up should allow me to execute it from within the installer. 
If lxc needs a zfs pool then that should be created by conjure-up, possibly showing me my current disk layout and letting me partition something for that use.

The same thing applies to every usecase that juju fulfills.. it really isnt useful for me to use conjure-up to install "something" and then end up realizing that some tiny config isnt quite right and I have to learn juju and lxc and lxd extensively after all.

https://www.stgraber.org/ has posted a guide on how to setup kubernetes locally with conjure-up.
Why does such a guide need to exist at all if this is supposedly a supported option in the installer? 
Something doesnt seem quite right here.

I know what you will say "well, we just provide a wrapper over juju, it's not our problem if something fails because juju/lxc/[...] is failing". Well then this will never be truly userfriendly because the problem ends up at my doorstep where I, as a user who *originally does not want to deal with juju or anything really* now has to deal with these other tools because conjure-up doesn't handle everything.

### LINKED ISSUE #596: provide way to de-conjure-up a spell
It would be nice to provide a way to de-conjure-up a spell. While 'juju destroy-model' works, it a) exposes Juju unnecessarily, and b) doesn't cover other things the spell might do to the host (e.g. the kubernetes-core spell adds a lxd profile)

battlemidget:
Fixes #596
Fixes #587 
Fixes #65 

image

image

### DIFF
--- a/conjureup/controllers/destroy/gui.py
+++ b/conjureup/controllers/destroy/gui.py
@@ -29,7 +29,7 @@ class Destroy:
                            cb=self.finish)
 
         app.ui.set_header(
-            title="Choose a Controller or Model to teardown",
+            title="Choose a deployment to teardown",
             excerpt=excerpt
         )
         app.ui.set_body(view)
--- a/conjureup/controllers/destroy/gui.py
+++ b/conjureup/controllers/destroy/gui.py
@@ -1,0 +1,38 @@
+from conjureup import controllers, juju
+from conjureup.app_config import app
+from conjureup.telemetry import track_exception, track_screen
+from conjureup.ui.views.destroy import DestroyView
+
+
+class Destroy:
+
+    def __init__(self):
+        self.view = None
+
+    def __handle_exception(self, exc):
+        track_exception(exc.args[0])
+        app.ui.show_exception_message(exc)
+
+    def finish(self, controller, model):
+        return controllers.use('destroyconfirm').render(controller, model)
+
+    def render(self):
+        models_map = {}
+        existing_controllers = juju.get_controllers()['controllers']
+        for cname, d in existing_controllers.items():
+            models_map[cname] = juju.get_models(cname)
+
+        track_screen("Destroy Controller")
+        excerpt = ("Press [ENTER] on the highlighted item to destroy")
+        view = DestroyView(app,
+                           models=models_map,
+                           cb=self.finish)
+
+        app.ui.set_header(
+            title="Choose a Controller or Model to teardown",
+            excerpt=excerpt
+        )
+        app.ui.set_body(view)
+
+
+_controller_class = Destroy
--- a/conjureup/controllers/destroyconfirm/gui.py
+++ b/conjureup/controllers/destroyconfirm/gui.py
@@ -12,12 +12,12 @@ class DestroyConfirm:
 
     def __handle_exception(self, exc):
         track_exception(exc.args[0])
-        app.ui.set_footer("Problem destroying the model")
+        app.ui.set_footer("Problem destroying the deployment")
         return app.ui.show_exception_message(exc)
 
     def __do_destroy(self, controller_name, model_name):
         track_event("Destroying controller", "Destroy", "")
-        app.ui.set_footer("Destroying {} model, please wait.".format(
+        app.ui.set_footer("Destroying {} deployment, please wait.".format(
             model_name))
         future = juju.destroy_model_async(controller=controller_name,
                                           model=model_name,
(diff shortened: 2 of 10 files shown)


battlemidget:
Please test with `conjure-down` and `conjure-down <controller> <model>` for tui version

conjureup/controllers/destroy/gui.py
```diff
@@ -0,0 +1,38 @@
...
+            models_map[cname] = juju.get_models(cname)
+
+        track_screen("Destroy Controller")
+        excerpt = ("Press [ENTER] on the highlighted item to destroy")
+        view = DestroyView(app,
+                           models=models_map,
+                           cb=self.finish)
+
+        app.ui.set_header(
+            title="Choose a Controller or Model to teardown",
```
mikemccracken: This probably shouldn't say "controller or model" since you can only tear down models.
Also I wonder if "model" is a juju vocab term that might be confusing.

conjureup/ui/views/destroy.py
```diff
@@ -0,0 +1,86 @@
...
+        total_items = []
+        for controller in self.controllers:
+            models = self.models[controller]['models']
+            if len(models) > 0:
+                total_items.append(Color.label(
+                    Text("{} ({})".format(controller,
+                                          models[0].get('cloud', "")))
+                ))
+                for model in models:
+                    label = "  {}".format(model['name'])
```
mikemccracken: we need a lot more info about the model here. At least the number of machines, and ideally a list of the apps that are deployed.

conjureup/ui/views/destroy_confirm.py
```diff
@@ -0,0 +1,85 @@
...
+        total_items.append(Text("Model name: {}".format(
+            self.model['name'])))
+        total_items.append(Text("Controller name: {}".format(
+            self.model['controller-name'])))
+        total_items.append(Text("Cloud: {}".format(
+            self.model['cloud'])))
+        total_items.append(Text("Status: {}".format(
+            self.model['status']['current'])))
+        total_items.append(Text("Online since: {}".format(
+            self.model['status']['since'])))
```
mikemccracken: This needs to have more info about the model. See the comment about the earlier page (further down the diff) - we should build on that info. Maybe we should just show the output of 'juju status', except for the relations.

The more info about what you're about to nuke, the better.

conjureup/controllers/destroy/gui.py
```diff
@@ -0,0 +1,38 @@
...
+        app.ui.show_exception_message(exc)
+
+    def finish(self, controller, model):
+        return controllers.use('destroyconfirm').render(controller, model)
+
+    def render(self):
+        models_map = {}
+        existing_controllers = juju.get_controllers()['controllers']
+        for cname, d in existing_controllers.items():
+            models_map[cname] = juju.get_models(cname)
```
mikemccracken: I think we should probably filter out the 'controller' model from the list of things you can kill.

If we want to enable killing off the whole controller, we should make that more obvious, and just use destroy-controller…

conjureup/ui/views/destroy.py
battlemidget: For now I added number of machines running and how long the model has been up

conjureup/ui/views/destroy_confirm.py
battlemidget: image

battlemidget: A good start here at least for this merge

conjureup/ui/views/destroy.py
battlemidget: image

conjureup/controllers/destroy/gui.py
battlemidget: A below screenshot shows the use of the word 'deployment' rather than controller or model

conjureup/ui/views/destroy_confirm.py
mikemccracken: Can we add machines count to the confirmation page too?

conjureup/controllers/destroyconfirm/gui.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(self):
+        self.view = None
+
+    def __handle_exception(self, exc):
+        track_exception(exc.args[0])
+        app.ui.set_footer("Problem destroying the deployment")
+        return app.ui.show_exception_message(exc)
+
+    def __do_destroy(self, controller_name, model_name):
+        track_event("Destroying controller", "Destroy", "")
```
mikemccracken: @battlemidget just noticed this, this event should be 'destroying model'

PR #638, opened 2017-01-30
Allow specifying model in headless mode

### LINKED ISSUE #635: Allow specifying model name in headless
We're using something like the following in headless:

```
conjure-up kubernetes-core localhost my-controller
conjure-up kubernetes-core localhost my-controller
```

But this creates model names that don't represent what the deployment is. Suggestion to allow a third positional argument to set model name:

```
conjure-up kubernetes-core localhost my-controller prod-test
conjure-up kubernetes-core localhost my-controller qa-cluster
```

Alternatively, it might be interesting to implicitly set the model name in the UI above the deploy screen as a prefilled but editable "Deployment name" to expose this functionality in the UI. I'm hesitant to put it there, choice fatigue, etc.

battlemidget:
Fixes #635

### DIFF
--- a/conjureup/app.py
+++ b/conjureup/app.py
@@ -90,6 +90,9 @@ def parse_options(argv):
     parser.add_argument('controller', nargs='?',
                         help="Name of a juju controller to target. "
                         "If not provided, a new one is created.")
+    parser.add_argument('model', nargs='?',
+                        help="Name of a juju model to target. "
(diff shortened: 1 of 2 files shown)


battlemidget:
I think since we're naming our models/controller to represent whats being deployed we can skip this

battlemidget:
this should do what marco wants

PR #660, opened 2017-02-08
Fixes #615

### LINKED ISSUE #615: num_units is not necessary with the architecture view
having num_units in the config view is vestigial from before the architecture view was around.

restricting users to num_units in the arch view is unnecessary and confusing (see #604 )

We should just remove num_units in config and let users do whatever they want in arch.

battlemidget:

### DIFF
--- a/conjureup/ui/views/applicationconfigure.py
+++ b/conjureup/ui/views/applicationconfigure.py
@@ -23,7 +23,6 @@ class ApplicationConfigureView(WidgetWrap):
         self.controller = controller
         self.application = application
         self.options_copy = self.application.options.copy()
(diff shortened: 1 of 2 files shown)


mikemccracken:
This isn't what #615 was about - that was saying that we don't need to be able to set num_units in the config screen. We probably do still want it shown in the application list view.

#615 just basically said that we should remove this widget: https://github.com/conjure-up/conjure-up/blob/master/conjureup/ui/views/applicationconfigure.py#L103

battlemidget:
ahh ok i was confused

battlemidget:
@mikemccracken lemme know how this updated PR looks

PR #663, opened 2017-02-08
Consolidate exception handler

battlemidget:
exception caller refactor to cleanup the duplicate functions we had laying around.

### DIFF
--- a/conjureup/controllers/deploy/gui.py
+++ b/conjureup/controllers/deploy/gui.py
@@ -9,12 +9,12 @@ from bundleplacer.assignmenttype import AssignmentType, atype_to_label
 from conjureup import async, controllers, juju, utils
 from conjureup.api.models import model_info
 from conjureup.app_config import app
+from conjureup.errors import handle_exception
 from conjureup.maas import setup_maas
 from conjureup.telemetry import track_event, track_exception, track_screen
 from conjureup.ui.views.app_architecture_view import AppArchitectureView
 from conjureup.ui.views.applicationconfigure import ApplicationConfigureView
 from conjureup.ui.views.applicationlist import ApplicationListView
-from ubuntui.ev import EventLoop
 
 DEPLOY_ASYNC_QUEUE = "DEPLOY_ASYNC_QUEUE"
 
@@ -46,17 +46,14 @@ class DeployController:
         else:
             self.sync_assignments()
 
-    def _handle_exception(self, tag, exc):
-        track_exception(exc.args[0])
-        app.ui.show_exception_message(exc)
-        self.showing_error = True
-        EventLoop.remove_alarms()
-
     def _pre_deploy_exec(self):
         """ runs pre deploy script if exists
         """
-        app.env['JUJU_PROVIDERTYPE'] = model_info(
-            app.current_model)['provider-type']
+        try:
+            app.env['JUJU_PROVIDERTYPE'] = model_info(
+                app.current_model)['provider-type']
+        except LookupError as e:
+            return handle_exception(e)
         app.env['JUJU_CONTROLLER'] = app.current_controller
         app.env['JUJU_MODEL'] = app.current_model
 
@@ -83,8 +80,7 @@ class DeployController:
         except AttributeError:
             result = json.loads(future.result())
         except:
-            return self._handle_exception(
-                'E003',
+            return handle_exception(
                 Exception(
                     "Problem with pre-deploy: \n{}, ".format(
                         future.result())))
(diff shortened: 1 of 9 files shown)


conjureup/controllers/deploy/gui.py
```diff
@@ -46,17 +46,14 @@ def init_machines_assignments(self):
...
     def _pre_deploy_exec(self):
         """ runs pre deploy script if exists
         """
-        app.env['JUJU_PROVIDERTYPE'] = model_info(
-            app.current_model)['provider-type']
+        try:
+            app.env['JUJU_PROVIDERTYPE'] = model_info(
+                app.current_model)['provider-type']
+        except LookupError as e:
+            return handle_exception(e)
```
mikemccracken: why only handle this specific exception? We probably need to stop if any other exception is raised also, right?

conjureup/controllers/steps/gui.py
```diff
@@ -27,14 +28,9 @@ def __init__(self):
...
 
-    def __handle_exception(self, tag, exc):
-        track_exception(exc.args[0], is_fatal=True)
-        EventLoop.remove_alarms()
-        app.ui.show_exception_message(exc)
-
     def get_result(self, future):
         if future.exception():
-            self.__handle_exception('E002', future.exception())
+            return handle_exception(future.exception())
```
mikemccracken: handle_exception doesn't return anything. it's clearer to write 
```
handle_exception(e)
return
```

conjureup/controllers/deploy/gui.py
```diff
@@ -335,14 +336,13 @@ def do_deploy_remaining(self):
...
     def finish(self):
         def enqueue_set_relations():
-            rel_future = juju.set_relations(self.applications,
-                                            app.ui.set_footer,
-                                            partial(self._handle_exception,
-                                                    "Error setting relations"))
+            rel_future = juju.set_relations(
+                self.applications,
+                app.ui.set_footer,
+                handle_exception)
```
mikemccracken: you still need partial here, as `handle_exception` requires an argument. might want to check for other places where you removed partial

conjureup/controllers/deploy/gui.py
```diff
@@ -335,14 +336,13 @@ def do_deploy_remaining(self):
...
             return rel_future
         f = async.submit(enqueue_set_relations,
-                         partial(self._handle_exception,
-                                 "Error setting relations"),
+                         handle_exception,
```
mikemccracken: still need partial

conjureup/controllers/deploy/gui.py
```diff
@@ -352,13 +352,10 @@ def enqueue_set_relations():
...
         track_screen("Deploy")
-        try:
-            future = async.submit(self._pre_deploy_exec,
-                                  partial(self._handle_exception, 'E003'),
-                                  queue_name=juju.JUJU_ASYNC_QUEUE)
-            future.add_done_callback(self._pre_deploy_done)
-        except Exception as e:
-            return self._handle_exception('E003', e)
+        future = async.submit(self._pre_deploy_exec,
+                              handle_exception,
```
mikemccracken: still need partial

conjureup/controllers/newcloud/gui.py
```diff
@@ -89,21 +86,21 @@ def __post_bootstrap_exec(self):
...
             except Exception as e:
-                return self.__handle_exception(e)
+                return handle_exception(e)
 
     def __post_bootstrap_done(self, future):
         try:
             result = json.loads(future.result().decode('utf8'))
         except Exception as e:
-            return self.__handle_exception(e)
+            return handle_exception(e)
```
mikemccracken: While we're changing this, we should really add more detail to this error specifically.
JSON parsing errors are terrible, they say shit like "expected : in line 1" when what we really want to see is the text it tried to parse, because it's usually a real error message that would've been helpful. Maybe instead of just passing `e` through, we could build a new exception that included the actual decoded result?

conjureup/errors.py
```diff
@@ -0,0 +1,7 @@
+from conjureup.app_config import app
+from conjureup.telemetry import track_exception
+
+
+def handle_exception(exc):
+    track_exception(exc.args[0])
+    return app.ui.show_exception_message(exc)
```
mikemccracken: handle_exception is replacing a bunch of different handlers that sometimes call `EventLoop.remove_alarms()` - should we be calling it here? If not, why not?

show_exception_message doesn't return anything, so we shouldn't return its return value (it'll be None, but this is unclear and we shouldn't make people go look)

battlemidget:
I'm not happy with this, going to close for now

PR #667, opened 2017-02-09
Better error dialog handling, especially for bootstrap failures

mikemccracken:
Several issues were causing problems with error handling in bootstrap.

an error message shown by the async code launched by newcloud could get overwritten by deploy.gui.render. 
worse, newcloud was calling deploy.gui.render twice, so there were extra chances for that to happen.

Most of this should be fixed.

Also fixes the spurious SystemExit traceback when the user hits 'OK' to exit in the error dialog.

### DIFF
--- a/conjureup/app_config.py
+++ b/conjureup/app_config.py
@@ -85,4 +85,8 @@ app = SimpleNamespace(
     headless=False,
 
     # Remote endpoint type (An enum, see download.py)
-    endpoint_type=None)
+    endpoint_type=None,
+
+    # Flag so we only show the initial cause even if future async
+    # events also raise cascading errors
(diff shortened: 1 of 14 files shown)


battlemidget:
Problem I'm having with this is once you press on 'localhost' and it goes to bootstrap it just hangs at the cloud selection screen and the logs indicate it is bootsrapping

battlemidget:
If i raise an exception inside the bootstrap function though the error view is seen correctly. Though the actual doing of the bootstrap causes the UI to freeze

battlemidget:
Tests and UI flow nicely now, exception error view is shown immediately with no additional async operations happening.

PR #677, opened 2017-02-15
Fixes an issue on localhost deploying to nested lxd

### LINKED ISSUE #676: headless conjure-up kubernetes-core fails, while GUI does work
root cause appears to be nested lxd containers,
but the GUI version does not nest the containers.

likely a difference in the paths used wrt the placement code

battlemidget:
Fixes #676


battlemidget:
Need to get this in for wider testing

PR #700, opened 2017-02-24
Allow for 'side-loading' additional charm options for applications

battlemidget:
In some cases like OpenStack there is a single bundle for multiple substrates,
like MAAS and LXD (localhost). This usually requires alterations of some of the
charm options like LXD to point to no block-devices if inside a container, etc.

This allows us to create a **bundle-custom.yaml** inside our spell with a dictionary of

```
services:
  lxd:
    options:
       block-devices: None
```

And that we can merge into the charm stores original bundle for further
customization by the spell authors.

### DIFF
--- a/conjureup/utils.py
+++ b/conjureup/utils.py
@@ -6,6 +6,7 @@ import pty
 import shutil
 import sys
 import uuid
+from collections import Mapping
 from subprocess import (
     DEVNULL,
     PIPE,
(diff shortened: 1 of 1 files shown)


conjureup/utils.py
```diff
@@ -319,7 +348,19 @@ def setup_metadata_controller():
...
-    bundle = Bundle(filename=bundle_filename)
+    # Load bundle data early so we can merge any additional charm options
+    with open(bundle_filename) as f:
+        bundle_data = yaml.load(f)
+
+    bundle_custom_filename = os.path.join(app.config['spell-dir'],
+                                          'bundle-custom.yaml')
+    if os.path.isfile(bundle_custom_filename):
+        with open(bundle_custom_filename) as f:
+            bundle_custom = yaml.load(f)
```
johnsca: I find [pathlib](https://docs.python.org/3/library/pathlib.html) to be a bit nicer, but that's just personal preference:

```python
bundle_custom_file = Path(app.config['spell-dir']) / 'bundle-custom.yaml'
if bundle_custom_file.is_file():
    bundle_custom = yaml.safe_load(bundle_custom_file.read_text())
    bundle_data = merge_dicts(bundle_data, bundle_custom)
```

PR #710, opened 2017-03-04
[wip] Switch from macumba to libjuju

johnsca:
This is not finished yet, but putting up as a pre-review.  It depends on PR juju/python-libjuju#63 which is also pending review.

The two main outstanding issues right now are:

* Headless deploys ghost mostly successfully but seems to drop the resource, and the charm goes into error, which conjure-up doesn't detect.
* Interactive mode seems to issue deploy commands but the applications don't get deployed.  I think the difference is the placement with the machine, so need to investigate that more.

I'm certain there are other issues that will come up once those are debugged a bit more, but I wanted to get this up in preparation for the sprint.


johnsca:
Also, I apparently don't have push access to this repo, so I had to push this to my fork.

battlemidget:
Thanks Cory I'll get you the right access tonight

battlemidget:
@johnsca I think you should have the right perms now for all the projects, can you see if you can push to a branch on this repo now?

johnsca:
Closing this to reopen from another branch

PR #751, opened 2017-03-21
Improve error handling around JaaS registration

### LINKED ISSUE #745: Improve handling of "cannot get user details" error during JAAS registration
While registering with JAAS, the user can get an error of the form:

```
Login failed, please try again: ERROR cannot get user details for "https://login.ubuntu.com/+id/xxxxxxx": not found
```

This doesn't seem to actually prevent the controller from being registered (although it might cause issues when using the controller, later), and is most likely related to the difference between oauth (command-line credentials) vs browser-based USSO login.

We should at a minimum detect this specific error and present the user with a more helpful flow that recommends they visit and login to https://jujucharms.com/ before proceeding.  In the end, this probably means that we need to use the browser-based USSO login method, possibly via the requests library if not through the user's own browser.

johnsca:
Fixes #745

### DIFF
--- a/conjureup/controllers/jaaslogin/gui.py
+++ b/conjureup/controllers/jaaslogin/gui.py
@@ -38,10 +38,13 @@ class JaaSLoginController:
                                       cb=self.register))
 
     def register(self, email=None, password=None, twofa=None):
-        juju.register_controller('jaas', JAAS_DOMAIN,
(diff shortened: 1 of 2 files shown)


battlemidget:
:+1:  once the tests pass

johnsca:
Accidentally committed a debug line.

PR #754, opened 2017-03-22
Fix handling of boolean step inputs

### LINKED ISSUE #733: strings that don't have a default specified -  didn't show an input box

### LINKED ISSUE #734: not able to use boolean values in steps
boolean values - I couldn't use them, encountered a parse error (expected str, bytes or os.PathLike object, not bool)

### LINKED ISSUE #735: password fields - inputs are shown in clear text

johnsca:
Fixes #733
Fixes #734
Fixes #735

Main issue for bools was that the values passed through `app.env` must be strings.

Fixed default handling for all input types (including `TypeError` for booleans, unmasked passwords, and missing fields with no defaults).

I also improved the error message in case a step had a yaml definition but no implementation.

### DIFF
--- a/conjureup/controllers/steps/common.py
+++ b/conjureup/controllers/steps/common.py
@@ -15,9 +15,9 @@ def set_env(inputs):
     for i in inputs:
         env_key = i['key'].upper()
         try:
-            input_key = i['input']
+            input_key = str(i['input'])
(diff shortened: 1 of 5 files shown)


johnsca:
I think this might also fix #733 but I have to test that

PR #765, opened 2017-03-24
Fixes #764

### LINKED ISSUE #764: can_sudo always fails
Running the following on a ubuntu image that doesn't have sudo password enabled:

```
ubuntu@tupac:~$ sudo -nv
sudo: a password is required
ubuntu@tupac:~$ sudo su 
root@tupac:/home/ubuntu# exit
ubuntu@tupac:~$ sudo -nv
sudo: a password is required
```

`sudo -nv` fails no matter what here. This problem is seen during a headless installation as well:

`conjure-up kubernetes-core localhost`

battlemidget:
Provide an additional sudo check for checking passwordless


battlemidget:
This passes my headless install test and successfully pulls kubectl from snap store without having to type a password in

PR #840, opened 2017-04-22
Add better add cloud

battlemidget:
Makes new clouds listed in clouds output for future use. Assigns credentials to
those clouds individually as well.

### DIFF
--- a/conjureup/controllers/newcloud/gui.py
+++ b/conjureup/controllers/newcloud/gui.py
@@ -71,7 +71,7 @@ class NewCloudController:
         region = None
         if credentials is not None:
             cloud_type = juju.get_cloud_types_by_name()[app.current_cloud]
-            if cloud_type == 'maas' or cloud_type == 'oracle':
+            if cloud_type == 'maas':
                 # Now that we are passed the selection of a cloud we create a
                 # new cloud name for the remainder of the deployment and make
                 # sure this cloud is saved for future use.
@@ -86,7 +86,8 @@ class NewCloudController:
                     juju.add_cloud(app.current_cloud,
                                    credentials.cloud_config(
                                        credentials.endpoint.value))
-
+            else:
+                common.save_creds(app.current_cloud, credentials)
         credentials_key = common.try_get_creds(app.current_cloud)
         app.loop.create_task(common.do_bootstrap(credentials_key,
                                                  region=region,
--- a/conjureup/controllers/newcloud/gui.py
+++ b/conjureup/controllers/newcloud/gui.py
@@ -70,32 +70,22 @@ class NewCloudController:
 
         region = None
         if credentials is not None:
-            common.save_creds(app.current_cloud, credentials)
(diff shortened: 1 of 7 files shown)


conjureup/controllers/newcloud/gui.py
```diff
@@ -70,32 +70,22 @@ def finish(self, credentials=None, back=False):
 
         region = None
         if credentials is not None:
```
battlemidget: Thinking maybe we should change the variable to 'provider'?

johnsca: Instead of `credentials`?  I think `credentials` is pretty accurate.  *shrug*

conjureup/controllers/newcloud/gui.py
```diff
@@ -70,32 +70,22 @@ def finish(self, credentials=None, back=False):
...
-                oracle_config = {
-                    'type': 'oracle',
-                            'description': 'Oracle Cloud',
-                            'auth-types': ['userpass'],
-                            'endpoint': endpoint,
-                            'regions': {
-                                'uscom-central-1': {}
-                            }
-                }
+            if cloud_type == 'maas' or cloud_type == 'oracle':
```
johnsca: Seeing this made me wonder what would happen if you selected an existing cloud of one of these types.  Following that code path, I noticed [newcloud/gui.py#L38](https://github.com/conjure-up/conjure-up/blob/master/conjureup/controllers/newcloud/gui.py#L38) and now I'm wondering why we're bothering to check the endpoint at all instead of just checking whether `juju.get_cloud(app.current_cloud)` returns something?  In particular, it seems like there's an edge case there where a given cloud type has existing credentials saved but no actual cloud (or the user is adding a second cloud of that type). **Edit:** I just realized that the only clouds we allow adding are the special-case ones, lxd (which follows an entirely separate branch), and maas & oracle (which both have a top-level `endpoint`, unlike all of the predefined clouds).  So I guess this concern is moot.

As for my original question, I think the flow here is ok, but I do think it might be easier to follow if the "register a new cloud" action was an entirely separate callback from "select an existing cloud" callback.  A refactor for another day, perhaps, unless you agree and feel like doing it as part of this.

battlemidget: I dont think this would be to hard to implement on this PR. I'll look at that soon

battlemidget: I'm going to do this refactor at a later time

conjureup/models/provider.py
```diff
@@ -116,10 +133,56 @@ def __init__(self):
...
-            self.address,
+            self.endpoint,
             self.apikey
         ]
 
+    def cloud_config(self, endpoint):
+        return {
+            'type': 'maas',
+            'auth-types': ['oauth1'],
+            'endpoint': endpoint
```
battlemidget: This could probably just be self.endpoint.value

Issue #545, opened 2017-01-09
look into dealing with existing credentials


battlemidget:
some cases you may have existing credentials in juju but want to create a new one or even select a different credential. should add a view to deal with this.

PR #847, opened 2017-04-27
integrate region selection in credentials view

### LINKED ISSUE #715: support bootstrap to a region for public cloud in gui mode
Allow selecting region in gui mode if applicable. This will require some additional UI widgets to allow a user to select from a list of regions, preferably showing up to the right of the cloud list and be specific to whatever cloud is highlighted.

battlemidget:
Fixes #715

image

### DIFF
--- a/conjureup/controllers/newcloud/common.py
+++ b/conjureup/controllers/newcloud/common.py
@@ -52,9 +52,9 @@ def try_get_creds(cloud):
         return existing_creds['credentials'][cloud]['default-credential']
 
     # XXX we should really prompt to select because this is non-deterministic
-    for k in existing_creds['credentials'][cloud].keys():
+    for k, v in existing_creds['credentials'][cloud].items():
         if 'default-region' in k:
-            continue
+            app.current_region = v
         else:
             return k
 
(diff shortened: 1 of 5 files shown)


battlemidget:
@johnsca Wanted to go ahead and get eyes on this. There is a little bit of work to not dimiss the credentials view if existing credentials exist. I would like to produce a selection of credentials the user can use otherwise prompt them for new credentials. In either case, however, I think we always want to show them a list of cloud regions to select from.

Thoughts?

conjureup/controllers/newcloud/common.py
```diff
@@ -52,9 +52,9 @@ def try_get_creds(cloud):
         return existing_creds['credentials'][cloud]['default-credential']
 
     # XXX we should really prompt to select because this is non-deterministic
```
battlemidget: I think this should be fixed in a future PR that adds the ability for users to select from a list of credentials for a cloud.

conjureup/controllers/newcloud/common.py
```diff
@@ -52,9 +52,9 @@ def try_get_creds(cloud):
...
-    for k in existing_creds['credentials'][cloud].keys():
+    for k, v in existing_creds['credentials'][cloud].items():
         if 'default-region' in k:
-            continue
+            app.current_region = v
```
battlemidget: Not perfect but should be addressed in a seperate PR concerning credential handling

conjureup/controllers/newcloud/common.py
```diff
@@ -178,14 +178,15 @@ def setup_lxdbr0_network():
...
     if not app.is_jaas:
         app.log.info('Bootstrapping Juju controller.')
         msg_cb('Bootstrapping Juju controller.')
         track_event("Juju Bootstrap", "Started", "")
         cloud_with_region = app.current_cloud
-        if region:
-            cloud_with_region = '/'.join([app.current_cloud, region])
+        if app.current_region:
+            cloud_with_region = '/'.join([app.current_cloud,
+                                          app.current_region])
```
battlemidget: This code also exists in `app.py`, should we move that code to the `tui.py` for this controller?

battlemidget:
> @johnsca Wanted to go ahead and get eyes on this. There is a little bit of work to not dimiss the credentials view if existing credentials exist. I would like to produce a selection of credentials the user can use otherwise prompt them for new credentials. In either case, however, I think we always want to show them a list of cloud regions to select from.

I realize this should've said in the next PR

johnsca:
Looks like #545 exists to track dealing with existing credentials.  Should we bump that up in priority?

PR #859, opened 2017-05-02
Safe load yaml

battlemidget:


battlemidget:
This problem still occurs even with safe_load, it looks like we'd have to ignore any type of tags (http://stackoverflow.com/questions/13280978/pyyaml-errors-on-in-a-string/13281292#13281292), I opened a bug on jujucharms so they can fix it on their end instead.

PR #870, opened 2017-05-09
Allow to run without installation

techtonik:


battlemidget:
Thanks for the PR, one question though, we do have `make dev` to handle this sort of thing. If you could let me know your thought process behind this so I can make a better decision I'd appreciate it.

techtonik:
`Makefile` doesn't work without `snapcraft`. Too much restrictions to run simple Python app.

battlemidget:
@techtonik fair enough, thanks!

PR #879, opened 2017-05-10
Fix JAAS registration for 2.2-beta4

### LINKED ISSUE #878: Juju beta4 breaks auth
Auth in juju 2.2-beta4 moves to per-controller macaroon files in `~/.local/share/juju/cookies/$controller.json`, rather than a single `~/.go-cookies` file.

This breaks the auth process for conjure-up, especially for JAAS.

johnsca:
Juju 2.2-beta4 switched to per-controller macaroon files.

Fixes #878

### DIFF
--- a/conjureup/juju.py
+++ b/conjureup/juju.py
@@ -190,50 +190,52 @@ async def bootstrap(controller, cloud, model='conjure-up', series="xenial",
 
 
 def has_jaas_auth():
-    oauth_token = Path('~/.local/share/juju/store-usso-token').expanduser()
-    go_cookies = Path('~/.go-cookies').expanduser()
-    if oauth_token.exists():
-        return True
-    if go_cookies.exists():
-        go_cookies = json.loads(go_cookies.read_text())
-        for cookie in go_cookies or []:
+    jaas_cookies = Path('~/.local/share/juju/cookies/jaas.json').expanduser()
+    if jaas_cookies.exists():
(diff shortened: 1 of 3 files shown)


johnsca:
This is blocked on juju/python-libjuju#124 but shouldn't require any further changes once that is fixed.

conjureup/juju.py
```diff
@@ -190,50 +190,52 @@ def add_config(k, v):
...
     try:
-        stdin = b''.join(b'%s\n' % bytes(f, 'utf8') for f in input)
+        stdin = b''.join(b'%s\n' % bytes(f, 'utf8')
+                         for f in [name, email, password, twofa])
         stdout, stderr = await asyncio.wait_for(proc.communicate(stdin),
                                                 timeout)
         stdout = stdout.decode('utf8')
         stderr = stderr.decode('utf8')
+
+        prefix = 'Enter a name for this controller: '
```
battlemidget: Do we provide any input for this or is it just us reading that prefix?

johnsca: We provide input for it, but Juju has started putting the prompt for it on stderr instead of stdout.  All of the other prompts go on stdout like they should.  We should probably file a bug, but it doesn't seem like it'd be high priority and the work-around isn't terrible.

battlemidget: Ok, sounds good

johnsca:
libjuju 0.4.2 has been released to pypi with the other part of the fix for this.

PR #887, opened 2017-05-15
Add screens for selecting region and credential

### LINKED ISSUE #882: region selection for jaas deployment
Need to verify that region selection is available via jaas deployments.

### LINKED ISSUE #883: make sure region selection is always visible
currently if a user credential is found the region selection is bypassed, we should make sure this is always seen

johnsca:
Fixes #882
Fixes #883


conjureup/controllers/bootstrap/common.py
```diff
@@ -0,0 +1,90 @@
...
+            cloud_with_region = app.current_cloud
+            if app.current_region:
+                cloud_with_region = '/'.join([app.current_cloud,
+                                              app.current_region])
+            success = await juju.bootstrap(app.current_controller,
+                                           cloud_with_region,
+                                           app.current_model,
+                                           credential=creds)
+            if not success:
+                pathbase = os.path.join(
```
battlemidget: Does it make sense to update these path creation to use Path instead? I think it would be nice to be able to assume that you can interact with any paths in the same fashion. This is just a question not something to do at the moment though

conjureup/controllers/bootstrap/gui.py
```diff
@@ -0,0 +1,6 @@
+from conjureup.app_config import app
+
+from . import common
+
+
+_controller_class = common.BootstrapController(app.ui.set_footer)
```
battlemidget: very nice, ty

conjureup/controllers/bootstrap/common.py
johnsca: Yeah, it would probably be a little bit cleaner.  This was just copypasta'd over, so I didn't go back and touch up everything I could have.  Something like:

```python
                log_file = '{}-bootstrap.err'.format(app.current_controller)
                log_file = Path(app.config['spell-dir']) / log_file
                err_log = log_file.read_text('utf8').splitlines()
```

battlemidget: Ok, as we iterate maybe do some incremental updates to switching over paths to using pathlib

conjureup/controllers/bootstrap/common.py
```diff
@@ -0,0 +1,86 @@
...
+                app.current_cloud,
+                utils.gen_hash())
+
+        if app.current_model is None:
+            app.current_model = utils.gen_model()
+
+        if not app.is_jaas:
+            await self.pre_bootstrap()
+            app.log.info('Bootstrapping Juju controller.')
+            self.msg_cb('Bootstrapping Juju controller.')
```
battlemidget: Not a blocker just a place holder, I think these should be replaced with self.emit

PR #911, opened 2017-06-03
Feature/910 enable redis state server

### LINKED ISSUE #168: make sharing info between steps easier
simplest way is to just expose a temp dir to every step and let them write files to communicate

### LINKED ISSUE #910: use state server for handling step results
Use redis backed storage for passing around step results

battlemidget:
This utilizes our StepModel in all run_step invocations. Also makes use of a special environment key 
https://github.com/conjure-up/conjure-up/wiki/Spell-Step-Specification#storing-step-results

You can test this by:

```
sudo apt install redis-server
make dev
. conjure-dev/bin/activate
conjure-up ~/local-path-to-spells/ghost
```

The ghost spell's step to set a url would look something like (step-01_set_url)

```
#!/bin/bash

set -eu

. $CONJURE_UP_SPELLSDIR/sdk/common.sh

juju config -m $JUJU_CONTROLLER:$JUJU_MODEL ghost url="$BLOGURL"

setResult "Visit your Ghost application at $BLOGURL"
exit 0
```

Fixes #910 
Fixes #168

### DIFF
--- a/conjureup/controllers/steps/common.py
+++ b/conjureup/controllers/steps/common.py
@@ -89,4 +89,8 @@ async def do_step(step_model, msg_cb):
     # Set environment variables so they can be accessed from the step scripts
     set_env(step_model.additional_input)
 
-    return await utils.run_step(step_model.name, step_model.title, msg_cb)
+    # Set result to be written to
+    app.env['RESULT'] = "conjure-up.{}.{}".format(app.config['spell'],
+                                                  step_model.result_key)
+
+    return await utils.run_step(step_model, msg_cb)
--- a/conjureup/controllers/steps/common.py
+++ b/conjureup/controllers/steps/common.py
@@ -89,8 +89,4 @@ async def do_step(step_model, msg_cb):
     # Set environment variables so they can be accessed from the step scripts
     set_env(step_model.additional_input)
 
-    # Set result to be written to
-    app.env['RESULT'] = "conjure-up.{}.{}".format(app.config['spell'],
-                                                  step_model.result_key)
-
     return await utils.run_step(step_model, msg_cb)
--- a/conjureup/models/step.py
+++ b/conjureup/models/step.py
@@ -35,4 +35,4 @@ class StepModel:
         return "<t: {} d: {} v: {} p:>".format(self.title,
                                                self.description,
                                                self.viewable,
-                                               self.path)
+                                               self.filename)
--- a/conjureup/models/step.py
+++ b/conjureup/models/step.py
@@ -5,15 +5,14 @@ from conjureup.app_config import app
 
 class StepModel:
 
-    def __init__(self, step, path, name):
+    def __init__(self, step, filename, name):
         self.title = step.get('title', '')
         self.description = step.get('description', '')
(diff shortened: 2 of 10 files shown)


conjureup/controllers/steps/common.py
```diff
@@ -89,4 +89,8 @@ def load_step(step_meta_path):
     # Set environment variables so they can be accessed from the step scripts
     set_env(step_model.additional_input)
 
-    return await utils.run_step(step_model.name, step_model.title, msg_cb)
+    # Set result to be written to
+    app.env['RESULT'] = "conjure-up.{}.{}".format(app.config['spell'],
```
johnsca: Do we want to provide a key prefix value as well, to make it easy for steps to namespace any data they want to store?  Could we handle the prefix automatically via the wrapper?  (If we do that, we could maybe document that the key is just `step_result` or such, without needing an env var for the key at all.)

We could use separate databases, but by default those are limited to 16 and we'd need to separately store the translations from spells to DB index.

conjureup/utils.py
```diff
@@ -136,19 +135,18 @@ def last_ten_lines(s):
...
             if proc.returncode != 0:
-                raise Exception("Failure in step {}".format(step_file))
+                raise Exception("Failure in step {}".format(step.path))
 
             if event_name is not None:
                 track_event(event_name, "Done", "")
 
-    return result if result else msg
+    result = app.state.get("conjure-up.{}.{}".format(app.config['spell'],
+                                                     step.result_key))
```
battlemidget: Probably support not having a message at all, less to confuse the user with. What do you think?

conjureup/controllers/steps/common.py
battlemidget: needs SPELL_NAME and STEP_NAME exported in environment

conjureup/models/step.py
```diff
@@ -5,14 +5,14 @@
...
-    def __init__(self, step, path, name):
+    def __init__(self, step, filename, name):
         self.title = step.get('title', '')
         self.description = step.get('description', '')
         self.result = ''
         self.viewable = step.get('viewable', False)
         self.needs_sudo = step.get('sudo', False)
         self.additional_input = step.get('additional-input', [])
-        self.path = path
+        self.filename = filename
```
johnsca: The `__repr__` function needs to be updated for this change.

conjureup/utils.py
```diff
@@ -136,19 +140,19 @@ def last_ten_lines(s):
...
                     async for line in f:
-                        if line.startswith('result:'):
-                            result = line
-                        msg = line
                         msg_cb(line)
                     await asyncio.sleep(0.01)
 
             if proc.returncode != 0:
-                raise Exception("Failure in step {}".format(step_file))
+                raise Exception("Failure in step {}".format(step.path))
```
johnsca: This should be `step.filename`

conjureup/models/step.py
johnsca: There's another reference in `conjureup/ui/widgets/step.py` as well.

PR #922, opened 2017-06-12
Make sure lxd profile is correct for our bridges

### LINKED ISSUE #921: lxdbr0 set as default for localhost cloud
conjure-up  2.2-rc2-20170612.1306
NovaLXD spell using localhost

+------------+----------+---------+---------+
|    NAME    |   TYPE   | MANAGED | USED BY |
+------------+----------+---------+---------+
| conjureup0 | bridge   | NO      | 0       |
+------------+----------+---------+---------+
| conjureup1 | bridge   | NO      | 0       |
+------------+----------+---------+---------+
| eno1       | physical | NO      | 0       |
+------------+----------+---------+---------+
| eno2       | physical | NO      | 0       |
+------------+----------+---------+---------+
| lxdbr0     | bridge   | YES     | 0       |
+------------+----------+---------+---------+

conjure-up.lxc profile show default
config: {}
description: Default LXD profile
devices:
  eth0:
    nictype: bridged
    parent: lxdbr0
    type: nic
  root:
    path: /
    pool: default
    type: disk
name: default
used_by: []

battlemidget:
Our internal lxd instance should only contain the bridge information for our
conjureup0 and conjureup1 network bridges to be used by conjure-up during
localhost deployments. In some cases the default profile was getting written
with lxdbr0 as it's parent bridge which causes issues as the certificates would
not match between bootstraps.

Fixes #921

### DIFF
--- a/conjureup/controllers/lxdsetup/common.py
+++ b/conjureup/controllers/lxdsetup/common.py
@@ -1,5 +1,7 @@
 import os
+import textwrap
 from pathlib import Path
+from tempfile import NamedTemporaryFile
 
 from conjureup import controllers, utils
 from conjureup.app_config import app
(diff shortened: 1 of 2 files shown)


battlemidget:
Talking with stgraber, all the config data is stored in the database and lxd would not read from 2 different ones. My thinking is maybe juju is accessing a lxd on a different socket (perhaps our internal lxd isn't up when juju runs bootstrap?) and pulls from whatever the host is and then writes the inital profile config based on that data. Could be a race condition, but, in the interest of making sure our profile is correct this will overwrite any pre-existing profile with the correct one during the initial setup.

PR #933, opened 2017-06-15
Add Sentry automatic error reporting

### LINKED ISSUE #931: Add Sentry support for tracking errors
We currently track errors with Google Analytics, but we need more details to help debug.  Sentry.io looks like a good fit.

Concerns:

* Ensure we don't capture credential data (mainly a concern from bootstrap logs)
* How much data / logs to capture?
* Should we prompt the user before attempting to send crash data?  (We don't for regular analytics data.)

johnsca:
Resolves #931

### DIFF
--- a/conjureup/app.py
+++ b/conjureup/app.py
@@ -3,8 +3,10 @@
 
 import argparse
 import asyncio
+import configparser
 import os
 import os.path as path
+import pathlib
 import platform
 import signal
 import subprocess
@@ -12,6 +14,7 @@ import sys
 import textwrap
 import uuid
 
+import raven
 import redis
 import yaml
 from prettytable import PrettyTable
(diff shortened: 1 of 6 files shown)


conjureup/app.py
```diff
@@ -63,6 +66,9 @@ def parse_options(argv):
                         help='Location of conjure-up managed spells directory',
                         default=os.path.expanduser(
                             "~/.cache/conjure-up-spells"))
+    parser.add_argument('--conf-file', dest='conf_file',
+                        help='Path to configuration file', type=pathlib.Path,
+                        default=pathlib.Path("~/.config/conjure-up.conf"))
```
battlemidget: note to self: make sure this is accessible within the snap as well

conjureup/app.py
```diff
@@ -90,6 +96,10 @@ def parse_options(argv):
                         dest='notrack',
                         help='Opt out of sending anonymous usage '
                         'information to Canonical.')
+    parser.add_argument('--noreport', action='store_true',
```
battlemidget: is this different from notrack?

conjureup/app.py
johnsca: It should be the same as the `.cache` folder, in that due to `--classic` it should be outside the snap.

PR #952, opened 2017-06-19
Remove dnsmasq hold on an interface during network creation

battlemidget:
In some cases, primarily during upgrades, dnsmasq will hold onto an interface
that needs to be managed by LXD. This causes issues during network creation
where it will give 'Address in use' even though the network is created at random.

https://sentry.io/canonical-pj/conjure-up/issues/296744258/

### DIFF
--- a/conjureup/controllers/lxdsetup/common.py
+++ b/conjureup/controllers/lxdsetup/common.py
@@ -7,6 +7,7 @@ from tempfile import NamedTemporaryFile
 
 from pkg_resources import parse_version
 
+import psutil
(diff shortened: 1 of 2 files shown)


afterbangx:
I'm currently having this issue. Is there a workaround until this fix is released?This is all new to me, I have a fresh install (installed today) of 16.04 and wanted to play around with OpenStack. Have been using the instructions  found [here](https://www.ubuntu.com/download/cloud/conjure-up) but it's failing every time with this error.

PR #976, opened 2017-06-22
Verify running user can access LXD daemon

### LINKED ISSUE #973: Warn user if not part of LXD group when deploying localhost
Some cases a user may not be part of the LXD group. Post a warning and a way to fix that prior to bootstrap so the user can re-start conjure-up.

battlemidget:
Make sure the user is in the LXD group, if group doesn't exist or user is not in
the group give them a clear exception message on how to fix. Additonally, these
errors do not need to be tracked as it is understood that users with improper
permissions will never be able to access the lxd daemon bundled or not.

Fixes #973

### DIFF
--- a/conjureup/events.py
+++ b/conjureup/events.py
@@ -8,11 +8,11 @@ from urwid import ExitMainLoop
 
 from conjureup import utils
 from conjureup.app_config import app
-from conjureup.telemetry import track_exception
 from conjureup.controllers.lxdsetup.common import (
-    LXDInvalidUserError,
-    LXDInvalidGroupError
+    LXDInvalidGroupError,
+    LXDInvalidUserError
 )
+from conjureup.telemetry import track_exception
 
 
 class Event(asyncio.Event):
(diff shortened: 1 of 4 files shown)


conjureup/events.py
```diff
@@ -119,7 +123,8 @@ def set(self, exit_code=None):
 # Keep a list of exceptions we know that shouldn't be logged
 # into sentry.
 NOTRACK_EXCEPTIONS = [
-    lambda exc: exc is OSError and exc.errno == errno.ENOSPC
+    lambda exc: exc is OSError and exc.errno == errno.ENOSPC,
+    lambda exc: exc is LXDInvalidUserError or exc is LXDInvalidGroupError
```
battlemidget: The report is still sent to sentry, not sure why @johnsca

johnsca: This should be:

```python
    lambda exc: isinstance(exc, (LXDInvalidUserError, LXDInvalidGroupError)),
```

battlemidget:
selection_010

snap/hooks/configure
```diff
@@ -48,6 +48,7 @@ if [[ -d $SNAP_COMMON/lxd ]]; then
 fi
 
 if [[ -f $SNAP_COMMON/lxd/unix.socket ]]; then
+    groupadd lxd || true
```
johnsca: Why are we only creating the group if the socket file already exists?  I would expect that if the socket file exists, there's a good chance that the group already exists anyway.  It seems like a fresh install is the case where we'd actually need to create the group, so this seems like it should be outside the `if`.

Issue #1009, opened 2017-06-26
Move steps view before deploy status


tvansteenburgh:
Change order.
Store step results.

johnsca:
![](https://github.trello.services/images/mini-trello-icon.png) [Move Steps view before Deploy](https://trello.com/c/n8vfimEl/6-move-steps-view-before-deploy)

battlemidget:
https://github.com/conjure-up/conjure-up/commit/8088792caedc8905d470e9aa691df435b890943e

PR #1037, opened 2017-07-14
Move Steps view before Deploy

johnsca:
Addresses #1009

The inputs for all of the post-deploy steps will be collected prior to the deploy screen, so that once a deployment is started, all of the interactive bits will be done and the user can just walk away.


johnsca:
From Travis:

```
  File "/opt/python/3.5.3/lib/python3.5/asyncio/tasks.py", line 239, in _step
    result = coro.send(None)
  File "/home/travis/build/conjure-up/conjure-up/conjureup/controllers/runsteps/tui.py", line 12, in run_steps
    for step in app.steps:
TypeError: 'NoneType' object is not iterable
```

johnsca:
Steps should be getting loaded [here](https://github.com/conjure-up/conjure-up/pull/1037/files#diff-ec3ed04dc18a4bc15b58f828912325b9R7).  There must be some flow that I haven't found that causes that view to be skipped.

conjureup/app.py
```diff
@@ -233,6 +228,15 @@ def main():
     app.state = redis.StrictRedis(host='localhost',
                                   port=opts.redis_port,
                                   db=0)
+    # confirm that the Redis connection is working
+    # XXX perhaps we should start the service if not running
```
battlemidget: Yea since we dont require root this could be easily handled in conjure-up

conjureup/app_config.py
```diff
@@ -122,10 +122,19 @@ class AppConfig:
     state = None
 
     # Sentry endpoint
-    sentry = None,
+    sentry = None
```
battlemidget: ugh nice catch that was my mistake still thinking it was a simplenamespace

conjureup/app.py
battlemidget: https://github.com/conjure-up/conjure-up/issues/1039

PR #1038, opened 2017-07-14
Fix Time

bondareko:


battlemidget:
This is a crazy amount of commits for this PR, please rebase from current master and re-apply your changes on that.

PR #1066, opened 2017-07-31
Enhanced VSphere support

battlemidget:
Visual Aid:

**Credential Screen**
conjure-up kubernetes-core_011

**VSphere region selection**
conjure-up ghost_014

**VSphere Setup Screen**

These properties are automatically queried from VSphere server. 

conjure-up kubernetes-core_013

### DIFF
--- a/conjureup/controllers/credentials/common.py
+++ b/conjureup/controllers/credentials/common.py
@@ -17,7 +17,4 @@ class BaseCredentialsController:
 
     def finish(self, cred):
         app.current_credential = cred
-        if app.current_cloud_type == 'localhost':
-            controllers.use('lxdsetup').render()
-        else:
-            controllers.use('controllerpicker').render()
+        controllers.use('regions').render()
--- a/conjureup/controllers/credentials/common.py
+++ b/conjureup/controllers/credentials/common.py
@@ -18,17 +18,6 @@ class BaseCredentialsController:
 
     def finish(self, cred):
         app.current_credential = cred
-
-        # We need to also attempt to login to the provider as we dont
-        # call save_credential.
-        # XXX: is this the best approach?
-        if app.current_cloud_type == 'vsphere' and \
-           not app.vsphere.authenticated:
-            vsphere_creds = juju.get_credential(app.current_cloud, cred)
-            vsphere_cloud = juju.get_cloud(app.current_cloud)
-            vsphere_provider = VSphere()
-            vsphere_provider.endpoint.value = vsphere_cloud['endpoint']
-            vsphere_provider.user.value = vsphere_creds['user']
-            vsphere_provider.password.value = vsphere_creds['password']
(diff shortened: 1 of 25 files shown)


conjureup/controllers/credentials/common.py
```diff
@@ -17,4 +18,17 @@ def __init__(self):
 
     def finish(self, cred):
         app.current_credential = cred
+
+        # We need to also attempt to login to the provider as we dont
+        # call save_credential.
+        # XXX: is this the best approach?
```
battlemidget: @johnsca im not sure I particularly like this approach but I couldnt come up with a more elegant way since each provider tends to provide different attributes.

conjureup/models/provider.py
```diff
@@ -445,19 +454,40 @@ def __init__(self):
...
             key='password'
         )
-        self.external_network = Field(
-            label='external network',
-            widget=StringEditor(),
-            key='external-network',
-            storable=False
-        )
+
+    def login(self):
```
battlemidget: maybe pass arguments here?

conjureup/models/credential.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def __init__(self, cloud, credential_name):
+        self.credential_name = credential_name
+        self.cloud = cloud
+        self.credential_obj = self._get_credential_object()
+
+    def get_cloud_type(self, name):
+        for name, cloud_type in get_cloud_types_by_name().items():
+            if name == self.cloud:
+                return cloud_type
```
johnsca: The `name` parameter is getting overwritten, there's no explicit fallback return value (with `None` being implicit), and overall this seems equivalent to `get_cloud_types_by_name().get(self.cloud)`

conjureup/models/credential.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def _get_credential_object(self):
+        cloud_type = self.get_cloud_type(self.cloud)
+        for credential in self.CREDENTIALS:
+            if credential.check_cloud_type(cloud_type):
+                return credential(self.cloud,
+                                  self.credential_name)
```
johnsca: What should we do if we don't find a match?

conjureup/models/provider.py
```diff
@@ -445,19 +460,37 @@ def __init__(self):
...
+
+        app.vsphere.client = VSphereClient(**credential)
+
+        try:
+            app.vsphere.client.login()
+            app.vsphere.authenticated = True
+        except:
+            raise Exception(
+                "Could not log in to VSphere, please "
+                "check your credentials and try again.")
```
johnsca: Another bare `except` and this also completely masks the original error.  We should at a minimum log what caused the failure, but we probably also want to distinguish between actual auth errors, connection errors, and other types of errors.

Issue #913, opened 2017-06-03
define an ordered list of controllers


battlemidget:
# Objective

This is to ease the maintenance burden when working with the ui controllers and
provide a mechanism for validating the current controllers in making sure it's
able to run.

# Rationale

It's a pain digging through the controller code to see which controllers are
previous and next in the flow of the installer. Having an ordered list of
controllers and a mechanism for automatically populating back()/forward()
methods for dealing with the navigation of those controllers.

Additionally, having checks in each controller to determine if they can actually
run or not will give us the ability to start conjure-up in different areas of
the install flow for things like resuming deployments, accessing the post
processing steps of existing deployments, and providing the first part in our
operations management feature for adding workloads on top of those existing
deployments if applicable.

battlemidget:
@johnsca this one might be a good first step to the operations management feature.

battlemidget:
@tvansteenburgh @johnsca After thinking about this more, it is more of a nice to have rather than something we actually need for our feature work this 2.3 cycle. I think we should put this particular feature on the back burner for now.

thoughts?

PR #1075, opened 2017-08-04
Provider refactor and add enhanced VSphere support

### LINKED ISSUE #1047: VSphere: Query API to gather network switches, ext network, datasources
Enhancement to give conjure-up ability to query vsphere api to get a list of virtual switches, external networks, datasources and present those to the user during credentials view. This will also need to be set via juju bootstrap-config so that the options persist to the bootstrap.

### LINKED ISSUE #1057: Add screen that lists network switch, external network, and data sources

### LINKED ISSUE #1058: Pass vsphere network parameters through to bootstrap

battlemidget:
Refactor:
    Refactored provider model to be the authoritative source when dealing with
    clouds, credentials, accessing api clients, and setting regions. This gives
    us the ability to make more provider specific assumptions when out of the
    realm of what juju can do.

    VSphere:
    This allows the user to select their external network, virtual switch, and
    datasource to use during the deployments.

    Fixes #1058
    Fixes #1057
    Fixes #1047

### DIFF
--- a/conjureup/app.py
+++ b/conjureup/app.py
@@ -374,8 +374,8 @@ def main():
                 parse_cli_cloud = app.argv.cloud.split('/')
                 cloud, region = parse_cli_cloud
                 app.log.debug(
-                    "Region found {} for cloud {}".format(app.cloud,
-                                                          app.region))
+                    "Region found {} for cloud {}".format(cloud,
+                                                          region))
             else:
                 cloud = app.argv.cloud
 
--- a/conjureup/app.py
+++ b/conjureup/app.py
@@ -36,6 +36,7 @@ from conjureup.download import (
     get_remote_url
 )
 from conjureup.log import setup_logging
+from conjureup.models.provider import SchemaErrorUnknownCloud, load_schema
 from conjureup.telemetry import SENTRY_DSN, track_event, track_screen
 from conjureup.ui import ConjureUI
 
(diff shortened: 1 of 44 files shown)


conjureup/controllers/regions/common.py
```diff
@@ -13,35 +13,42 @@ def __init__(self):
...
     def finish(self, region):
-        app.current_region = region
-        controllers.use('credentials').render()
+        app.provider.region = region
+        if app.provider.cloud_type == 'localhost':
+            controllers.use('lxdsetup').render()
+        elif app.provider.cloud_type == 'vsphere':
+            controllers.use('vspheresetup').render()
+        else:
+            controllers.use('controllerpicker').render()
```
battlemidget: Also thoughts on condensing this up? We already have the provider loaded so going through and requerying seems unnecessary now

johnsca: Well, this touches on #913 and part of what I'd like to see for that is for every controller to be able to define a "skip me if this is true" check so that next and back could always just go through the list of controllers, starting with the current one, and figure out which one should be shown.  Essentially, formalizing the "journey" doc.

conjureup/app.py
```diff
@@ -383,24 +368,36 @@ def main():
...
+            region = None
             if '/' in app.argv.cloud:
                 parse_cli_cloud = app.argv.cloud.split('/')
-                app.current_cloud, app.current_region = parse_cli_cloud
+                cloud, region = parse_cli_cloud
                 app.log.debug(
-                    "Region found {} for cloud {}".format(app.current_cloud,
-                                                          app.current_region))
+                    "Region found {} for cloud {}".format(app.cloud,
+                                                          app.region))
```
johnsca: These shouldn't be prefixed with `app.`

conjureup/controllers/regions/common.py
battlemidget: Ok, will leave it for now and revisit when when work on the controller navigation

johnsca:
I added a commit to fix a `NoneType` error due to the regions list not being populated when entering a new VSphere cloud.  Other than that, it works as expected for both new and saved VSphere clouds and the refactors look good.  +1 to merge

PR #1112, opened 2017-08-23
Checks for availability of LXD

### LINKED ISSUE #1049: Add check for LXD presence
LXD is optional dependency, we'll check for existence of snap LXD and alert the user if they want to select localhost as their provider. We can also put in a task to continually check for LXD's presence if the user decides to perform `snap install lxd` in another window.

battlemidget:
This will check for LXD existence, and update the UI accordingly.

To test:

```
sudo snap remove lxd
sudo apt-get remove -qy lxd lxd-client

conjure-up
```
Once you get to the cloud list you should see localhost greyed out with a
message about how to install snap version of lxd.

next, in another terminal (while leaving conjure-up running)

```
sudo snap install lxd --edge # For now until 2.17 is released.
```

Watch conjure-up update the localhost selection by re-enabling it and allowing
you to deploy to it.

Also, a lot of cleanup removing anything LXD related that we customized
ourselves. This paves the way for the next set of patches that allow the user to
further customize how they want their localhost provider to be configured for
their deployment.

@johnsca please pay closer attention to the async code as I want to make sure it is
keeping with the style of the rest of our async related code.

Fixes #1049


conjureup/lxd.py
```diff
@@ -0,0 +1,86 @@
...
+class LXDClientError(Exception):
+    pass
+
+
+class LXDClientJSONError(Exception):
+    pass
+
+
+class LXDClient:
+    """ Simple wrapper around lxc query or the older curl option for querying
```
battlemidget: need to update comment

conjureup/lxd.py
```diff
@@ -0,0 +1,86 @@
...
+    LXD API server
+    """
+
+    def __init__(self):
+        self.minimum_support_version = parse_version('2.16')
```
battlemidget: needs to be updated once 2.17 is released, hopefully tomorrow

conjureup/lxd.py
```diff
@@ -0,0 +1,86 @@
...
+        been met
+        """
+        try:
+            return await self.query()
+        except (LXDClientJSONError, LXDClientError):
+            if self.retry_query_count == 10:
+                raise
+            self._set_lxd_dir_env()
+            self.retry_query_count += 1
+            asyncio.sleep(2)
```
battlemidget: can probably remove this sleep as we use a sleep in the views/cloud.py

conjureup/ui/views/cloud.py
```diff
@@ -56,46 +64,113 @@ def _build_footer(self):
         ])
         return footer_pile
 
+    async def _monitor_localhost(self):
```
battlemidget: i could potentially move this into the controller code, @johnsca thoughts?

conjureup/controllers/clouds/gui.py
```diff
@@ -53,16 +55,27 @@ def render(self):
...
+
+        self.view = CloudView(app,
+                              public_clouds,
+                              custom_clouds,
+                              cb=self.finish)
+
+        if 'localhost' in cloud_types:
+            # Need to load the localhost provider here early so our LXD methods
+            # are available
+            app.provider = load_schema('localhost')
```
johnsca: This seems fragile.  Why not make the `LocalhostProvider` instance a param to the task, rather than sticking it in the global state?

conjureup/controllers/clouds/common.py
```diff
@@ -0,0 +1,38 @@
...
+            self.retry_count = 0
+            cb()
+        except (LocalhostError,
+                LocalhostJSONError):
+            if self.retry_count == self.max_retry:
+                raise
+            app.provider._set_lxd_dir_env()
+            self.retry_count += 1
+            await asyncio.sleep(2)
+            await self._monitor_localhost(cb)
```
johnsca: Recursion seems like an odd way to do a retry loop.  I think a `for` loop would be easier to follow (it took me a second to figure out that this was a retry loop).  Actually, why even have a max retry?  If the user sits on the cloud selection screen for too long it would just stop working.  I think a cancel event (using `run_with_interrupt`) would be better, so that the task can just keep trying for as long as needed and then be cancelled cleanly when a cloud is selected.  Something like:

```python
while not self.cancel_monitor.is_set():
    try:
        await app.provider.is_server_available()
        events.LXDAvailable.set()
        cb()
        return
    except (LocalhostError, LocalhostJSONError):
        await utils.run_with_interrupt(asyncio.sleep(2), self.cancel_monitor)
    except FileNotFoundError:
        if app.headless:
            utils.error("Unable to find lxc executable, please make "
                        "sure LXD is installed: `sudo snap install lxd`")
            events.Shutdown.set(1)
        await utils.run_with_interrupt(asyncio.sleep(5), self.cancel_monitor)
```

conjureup/controllers/clouds/tui.py
```diff
@@ -18,7 +20,10 @@ def finish(self):
     def render(self):
         utils.info(
             "Summoning {} to {}".format(app.argv.spell, app.provider.cloud))
-        self.finish()
+        if app.provider.cloud_type == cloud_types.LOCALHOST:
+            app.loop.create_task(self._monitor_localhost(self.finish))
```
johnsca: I don't understand why we're doing this task on TUI, since it's non-interactive?  Why not just fail fast if lxd isn't already available?

conjureup/controllers/lxdsetup/gui.py
```diff
@@ -1,16 +1,9 @@
...
-
 from . import common
 
 
 class LXDSetupController(common.BaseLXDSetupController):
     def render(self):
-        if len(self.ifaces) == 1:
-            return self.setup(self.ifaces[0])
-
-        view = LXDSetupView(app.provider, self.setup)
```
johnsca: Why aren't we showing the setup view now?  Does it need more work before it's useful?

conjureup/juju.py
```diff
@@ -182,7 +182,8 @@ def add_model_defaults(k, v):
         with open(pathbase + ".err", 'w') as errf:
             proc = await asyncio.create_subprocess_exec(*cmd,
                                                         stdout=outf,
-                                                        stderr=errf)
+                                                        stderr=errf,
+                                                        env=app.env)
```
johnsca: Nice catch :+1:

conjureup/controllers/lxdsetup/gui.py
battlemidget: Yea this need the next phase of the code for selecting/configure network and storage pools

PR #1119, opened 2017-08-28
Fix subordinates not showing on Deploy Status

johnsca:
In #1042, placeholders were added to the Deploy Status screen to avoid it being empty when first shown.  However, since the placeholders are based on the expected number of units, which for subordinates is 0,
placeholders can not be created for subordinates, and the view didn't account for them being added later.  They will now be added as they appear in the model.


johnsca:
[(2) show subordinates in deploy status view](https://trello.com/c/tfjbL42Z/42-2-show-subordinates-in-deploy-status-view)

PR #1136, opened 2017-09-05
Use cloud white/blacklist from selected addons only

johnsca:
The cloud filtering from the addons was too aggressive, limiting by all
available addons for the selected spell rather than just the chosen
addons.


johnsca:
[bug: cloud list inaccurate for spells with add-ons](https://trello.com/c/qIqdGsHL/63-bug-cloud-list-inaccurate-for-spells-with-add-ons)

PR #1146, opened 2017-09-08
Add support for `conjure-up $ADDON`

### LINKED ISSUE #1139: conjure-up $ADDON
Provide ability for users to skip spell/addon selections with aliases to known addons+spell combinations.

We would define an aliases file in the registry that mapped addon aliases to preferred spells, ie `conjure-up deis` would deploy Canonical Distribution of Kubernetes to run Deis on.

battlemidget:
Fixes #1139

### DIFF
--- a/conjureup/app.py
+++ b/conjureup/app.py
@@ -139,6 +139,8 @@ async def _start(*args, **kwargs):
 
     if app.headless:
         controllers.use('clouds').render()
+    elif app.selected_addons:
+        controllers.use('clouds').render()
     else:
         controllers.use('addons').render()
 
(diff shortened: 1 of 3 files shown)


conjureup/app.py
```diff
@@ -274,10 +276,29 @@ def main():
     with open(spells_index_path) as fp:
         app.spells_index = yaml.safe_load(fp.read())
 
+    addons_aliases_index_path = os.path.join(app.config['spells-dir'],
+                                             'addons-aliases.yaml')
+    with open(addons_aliases_index_path) as fp:
+        app.addons_aliases = yaml.safe_load(fp.read())
```
johnsca: This should probably handle ENOENT errors (or check for existence) on the off chance that someone uses this with a spells repo that doesn't have any aliases.

PR #1181, opened 2017-10-05
Add support for bundle overrides from CLI for testing

johnsca:


johnsca:
[(0.5) cdoqa req: conjure-up --bundle-fragment, provide bundle overrides via cli](https://trello.com/c/UTX5XokM/73-05-cdoqa-req-conjure-up-bundle-fragment-provide-bundle-overrides-via-cli)

PR #1183, opened 2017-10-10
Re-add openstack as a provider option

### LINKED ISSUE #1171: Add ability to add new OpenStack cloud via conjure-up
conjure-up currently allows for registering new localhost, MAAS, and vSphere clouds, but not OpenStack.  We should add support for registering a new OpenStack cloud.

battlemidget:
This was missed during some of the refactoring, adding this option back in so
custom openstack clouds can be utilized.

Fixes #1171

### DIFF
--- a/conjureup/consts.py
+++ b/conjureup/consts.py
@@ -4,7 +4,7 @@ UNSPECIFIED_SPELL = '_unspecified_spell'
 JAAS_CLOUDS = {'ec2', 'azure', 'gce'}
 JAAS_DOMAIN = 'jimm.jujucharms.com'
 JAAS_ENDPOINT = JAAS_DOMAIN + ':443'
-PROVIDER_TYPES = ['localhost', 'maas', 'vsphere', 'openstack']
+CUSTOM_PROVIDERS = ['localhost', 'maas', 'vsphere', 'openstack']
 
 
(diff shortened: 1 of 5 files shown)


conjureup/controllers/clouds/gui.py
```diff
@@ -18,7 +18,7 @@ def finish(self, cloud):
         """
         self.cancel_monitor.set()
 
-        if cloud in ['localhost', 'vsphere', 'maas']:
+        if cloud in ['localhost', 'vsphere', 'maas', 'openstack']:
```
johnsca: This should use the new list in consts.

conjureup/consts.py
```diff
@@ -4,6 +4,7 @@
 JAAS_CLOUDS = {'ec2', 'azure', 'gce'}
 JAAS_DOMAIN = 'jimm.jujucharms.com'
 JAAS_ENDPOINT = JAAS_DOMAIN + ':443'
+PROVIDER_TYPES = ['localhost', 'maas', 'vsphere', 'openstack']
```
johnsca: I'm not sure this is the best name, since it is a bit confusing with the list of Provider models.  Maybe something like `CUSTOM_CLOUDS`?

PR #1217, opened 2017-11-03
Differentiate DeploymentFailure error

johnsca:
Make DeploymentFailures their own exception type so they're easier to find in Sentry.

Also remove retry logic that was recently added back in to wait_for_applications because it's handled in juju-wait.


battlemidget:
I'll fix the isort issue when I get back tonight

PR #1234, opened 2017-11-17
Replace charmhelpers with kv

### LINKED ISSUE #1231: Replace charmhelpers with kv
The only thing in charmhelpers we use is the unitdata, making it an unnecessarily heavy-weight dependency.  The [kv](https://pypi.python.org/pypi/kv/) library does what we need and only what we need, so we should switch to that instead.

johnsca:
The charmhelpers library contains a lot more code than we need, since we only use the sqlite-backed KV store, and it doesn't work on Mac OS.

Fixes #1231

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,4 +30,4 @@ script:
 after_script:
   - cat ~/.cache/conjure-up/conjure-up.log
   - cat ~/.cache/conjure-up/ghost/*.err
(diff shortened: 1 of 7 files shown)


johnsca:
The test failure is because it's not using the updated spells branch (see PR conjure-up/spells#117).  Testing locally with that branch works.

PR #1285, opened 2017-12-22
Remove placement restrictions on localhost

battlemidget:

### DIFF
--- a/conjureup/controllers/configapps/gui.py
+++ b/conjureup/controllers/configapps/gui.py
@@ -38,7 +38,7 @@ class ConfigAppsController:
         """
         bundle = app.metadata_controller.bundle
 
-        if len(bundle.machines) == 0 or app.provider.cloud == "localhost":
+        if len(bundle.machines) == 0:
             self.generate_juju_machines()
         else:
             self.sync_assignments()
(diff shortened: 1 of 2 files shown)


conjureup/controllers/configapps/gui.py
```diff
@@ -67,16 +67,11 @@ def generate_juju_machines(self):
...
-                if app.provider.cloud == "localhost":
-                    app.log.info("Ignoring placement spec because we are "
-                                 "deploying to LXD: {}".format(
-                                     bundle_application.placement_spec))
-                else:
-                    app.log.warning("Ignoring placement spec because no "
-                                    "machines were set in the "
-                                    "bundle: {}".format(
-                                        bundle_application.placement_spec))
+            if not bundle_application.placement_spec:
```
johnsca: The `not` shouldn't be there, but I think we might as well just remove the entire warning.  I'm not even convinced it's true, anyway, since we don't actually strip out the placement directive and there's no real reason you couldn't say "put these two machines on 0" and have that just refer to the first automagically created machine.

conjureup/controllers/deploy/common.py
```diff
@@ -22,10 +22,7 @@
                                           msg_cb=msg_cb)
     tasks = []
     for service in applications:
-        if cloud_types[app.provider.cloud] == "localhost":
-            # ignore placement when deploying to localhost
-            service.placement_spec = None
-        elif service.placement_spec:
+        if service.placement_spec:
```
johnsca: Gotta remove `cloud_types` now.

conjureup/controllers/deploy/common.py
```diff
@@ -22,10 +22,7 @@
                                           msg_cb=msg_cb)
     tasks = []
     for service in applications:
-        if cloud_types[app.provider.cloud] == "localhost":
```
johnsca: Gotta remove `cloud_types` now.

PR #1296, opened 2018-01-11
Fix addon steps not being run

### LINKED ISSUE #1295: addon steps dont seem to be executing
I'm working on moving logging and metric steps to addons for cdk:

https://github.com/kwmonroe/spells/tree/feature/cdk-log-metric-addons/canonical-kubernetes/addons

Using that repo as my `--spells-dir` and conjure-up snap from beta:

```
tracking:    beta
installed:   2.5.1-20180106.0201 (919) 91MB classic
refreshed:   2018-01-05 20:45:25 -0600 CST
```

I select CDK, and then *all* the addons.  It doesn't seem like the `./addons/*/steps/*/*` bits are being processed.  I don't see any UI to prompt for input like this:

https://github.com/conjure-up/spells/blob/master/canonical-kubernetes/addons/deis/steps/01_install-workflow/metadata.yaml#L7

Nor is my deployment configured per `after-deploy` like this:

https://github.com/kwmonroe/spells/blob/feature/cdk-log-metric-addons/canonical-kubernetes/addons/graylog/steps/01_install-graylog/after-deploy

When I run through a deployment with my spells repo, the final CDK screen doesn't mention helm, prometheus, deis, or graylog at all, further making me think perhaps the `setResult` bits in `after-deploy` aren't running:

screen shot 2018-01-10 at 1 57 11 pm

johnsca:
Addons were missed during a refactor of the storage location for steps.

Fixes #1295

### DIFF
--- a/conjureup/models/addon.py
+++ b/conjureup/models/addon.py
@@ -32,9 +32,12 @@ class AddonModel:
         self.path = Path(app.config['spell-dir']) / 'addons' / name
         self.metadata = self._read('metadata.yaml')
         self.bundle = self._read('bundle.yaml')
-        self.steps = [StepModel.load(step_path)
+        self.friendly_name = self.metadata['friendly-name']
+        self.steps = [StepModel.load(step_path,
+                                     source=self.friendly_name)
                       for step_path in
-                      sorted((self.path / 'steps').glob('*.yaml'))]
+                      sorted((self.path / 'steps').glob('*'))
+                      if step_path.is_dir()]
 
     def _read(self, filename):
         filepath = self.path / filename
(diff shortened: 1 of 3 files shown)


conjureup/models/addon.py
```diff
@@ -32,9 +32,12 @@ def __init__(self, name):
         self.path = Path(app.config['spell-dir']) / 'addons' / name
         self.metadata = self._read('metadata.yaml')
         self.bundle = self._read('bundle.yaml')
-        self.steps = [StepModel.load(step_path)
+        self.friendly_name = self.metadata['friendly-name']
+        self.steps = [StepModel.load(step_path,
+                                     source=self.friendly_name)
```
kwmonroe: wat? why is the source arg the friendly name?

johnsca: It shows up in some error messages.  I started out using "spell" or the short name of the addon, which is fine for logging but I thought it made the error messages less helpful.  I'm fine changing it to the short name.

PR #1297, opened 2018-01-11
Remove attribute duplicating property

johnsca:


johnsca:
Test failure is spurious (some sort of issue with lxd on travis)

PR #1318, opened 2018-01-26
Convert Deploy Status screen to BaseView to enable scrolling

johnsca:
Also fixed an issue tracking screens with dynamic titles (i.e., ones where the title contains the spell or application name) so that they're consistent and we can get good metrics.

I did see some oddness in the rendering of the scrollbar on the Deploy Status screen, but that might have just been my terminal, and it didn't affect the functionality in any way.

### DIFF
--- a/conjureup/ui/views/deploystatus.py
+++ b/conjureup/ui/views/deploystatus.py
@@ -1,19 +1,49 @@
 import random
 
-from ubuntui.utils import Color, Padding
+from ubuntui.utils import Color
+from ubuntui.widgets.hr import HR
 from ubuntui.widgets.juju.service import ServiceWidget
 from ubuntui.widgets.table import Table
-from urwid import Text, WidgetWrap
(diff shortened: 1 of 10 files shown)


conjureup/ui/views/deploystatus.py
```diff
@@ -1,19 +1,50 @@
...
+from conjureup.ui.views.base import BaseView
 
-class DeployStatusView(WidgetWrap):
 
-    def __init__(self, app):
-        self.app = app
+class PileTable(Table):
+    # ubuntui's Table uses a ListBox, which is a box widget and thus
+    # incompatible with Scrollable, so this uses a Pile instead.
+    # TODO: Fix or add this in ubuntui
```
battlemidget: ack, would like to see this in ubuntui

battlemidget:
I couldn't reproduce the deploy status weirdness you were seeing

PR #1342, opened 2018-02-23
Check type of default value for step fields

### LINKED ISSUE #1339: Handle type mismatches between input field and default values better
It looks like if the type of a step field doesn't match the type of the default value, an exception will happen:

```
 File "/snap/conjure-up/967/lib/python3.6/site-packages/conjureup/controllers/showsteps/gui.py", line 33, in show_steps
   await self.show_step(step)
 File "/snap/conjure-up/967/lib/python3.6/site-packages/conjureup/controllers/showsteps/gui.py", line 38, in show_step
   step_widget = StepForm(app, step)
 File "/snap/conjure-up/967/lib/python3.6/site-packages/conjureup/ui/widgets/step.py", line 51, in __init__
   self._build_form_fields())
 File "/snap/conjure-up/967/lib/python3.6/site-packages/conjureup/ui/widgets/step.py", line 174, in _build_form_fields
   input_type(default=value), i['type'])
 File "/snap/conjure-up/967/lib/python3.6/site-packages/ubuntui/widgets/input.py", line 28, in __init__
   self._edit.set_edit_text(default)
 File "/snap/conjure-up/967/lib/python3.6/site-packages/urwid/widget.py", line 1354, in set_edit_text
   text = self._normalize_to_caption(text)
 File "/snap/conjure-up/967/lib/python3.6/site-packages/urwid/widget.py", line 1414, in _normalize_to_caption
   return text.decode('ascii')
AttributeError: 'bool' object has no attribute 'decode'
2018-02-22 21:58:03,772 [DEBUG] conjure-up/canonical-kubernetes - __init__.py:29 - Showing dialog for exception: 'bool' object has no attribute 'decode'
2018-02-22 21:58:03,884 [ERROR] conjure-up/canonical-kubernetes - sentry.errors.serializer: manager.py:78 - 'StepForm' object has no attribute '_contents'
```

Sentry report: https://sentry.io/canonical-pj/conjure-up/issues/471077427/

johnsca:
Fixes #1339

### DIFF
--- a/conjureup/ui/widgets/step.py
+++ b/conjureup/ui/widgets/step.py
@@ -170,6 +170,15 @@ class StepForm(Pile):
                     select_w.select_option(i['default'])
                     field = StepField(key, label, select_w, i['type'])
                 else:
+                    is_yesno = input_type is YesNo
+                    is_intedit = input_type is IntegerEditor
+                    if value is not None and (
+                        (isinstance(value, bool) and not is_yesno) or
+                        (isinstance(value, int) and not is_intedit) or
+                        isinstance(value, float)  # floats must be strings
+                    ):
+                        raise ValueError('Invalid default value for field: '
+                                         '{}'.format(label))
                     field = StepField(key, label,
                                       input_type(default=value), i['type'])
                 self.fields.append(field)


conjureup/ui/widgets/step.py
```diff
@@ -170,6 +170,14 @@ def _build_form_fields(self):
                     select_w.select_option(i['default'])
                     field = StepField(key, label, select_w, i['type'])
                 else:
+                    is_yesno = input_type is YesNo
+                    is_intedit = input_type is IntegerEditor
+                    if value is not None and (
+                        (isinstance(value, bool) and not is_yesno) or
+                        (isinstance(value, int) and not is_intedit)
+                    ):
```
Cynerva: Do we need to worry about other types here? e.g. `default=2.5` resulting in a `float`?

That's my only concern, otherwise this looks good.

johnsca: Float values don't have their own input type, so I'm assuming we don't support them at all and they'd be to be configured as `type: string` and `default: "2.5"`.

Cynerva: Aye, `default: 2.5` is not supported, but we're trying to provide a more descriptive user when the user does something wrong, right?

If I'm reading this right, if they mistakenly put `default: 2.5`, then `value` will be of type `float` and this if condition will evaluate to `False`. Possibly leading to an unfriendly error like `'float' object has no attribute 'decode'` or whatever.

PR #1349, opened 2018-02-28
Put charmstore in app_config, update application configure view

battlemidget:

### DIFF
--- a/conjureup/app.py
+++ b/conjureup/app.py
@@ -16,6 +16,7 @@ import uuid
 
 import raven
 import yaml
+from juju.model import CharmStore
 from kv import KV
 from prettytable import PrettyTable
 from raven.transport.requests import RequestsHTTPTransport
@@ -406,6 +407,9 @@ def main():
 
     app.loop = asyncio.get_event_loop()
     app.loop.add_signal_handler(signal.SIGINT, events.Shutdown.set)
+
+    # Enable charmstore querying
+    app.juju.charmstore = CharmStore(app.loop)
     try:
         if app.argv.cloud:
             cloud = None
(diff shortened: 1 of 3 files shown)


conjureup/ui/views/applicationconfigure.py
```diff
@@ -25,45 +25,50 @@ def __init__(self, application, close_cb):
         super().__init__()
 
     def build_widget(self):
+        app.loop.create_task(self._build_widget())
+        return []
```
battlemidget: Do i need this return here still?

johnsca: Yes.  `build_widget` is expected to return either a widget or a list that will be put into a `Pile`.  If you don't have a return, `None` is implied and it leads to the `NoneType has no attribute rows` error.

conjureup/app.py
```diff
@@ -16,6 +16,7 @@
 
 import raven
 import yaml
+from juju.model import CharmStore
```
johnsca: Really need to refactor this in python-libjuju.  `juju.model` is the wrong place for it.  Ah well.

conjureup/ui/views/applicationconfigure.py
```diff
@@ -25,45 +25,50 @@ def __init__(self, application, close_cb):
...
                                    value_changed_callback=self.handle_scale)
         ws.append(num_unit_ow)
-        ws += self.get_whitelisted_option_widgets()
+        ws += await self.get_whitelisted_option_widgets()
         self.toggle_show_all_button_index = len(ws) + 1
         self.toggle_show_all_button = SecondaryButton(
             "Show Advanced Configuration",
-            self.do_toggle_show_all_config)
+            lambda sender: app.loop.create_task(
+                self.do_toggle_show_all_config()))
```
johnsca: Might be nice to add some logic to `SecondaryButton` to detect coroutines and do the right thing, so that we don't have to deal with this all over the place.

PR #1361, opened 2018-03-09
Clean up application list view

battlemidget:
This removes all the bundleplacer assignment/sync code that won't be necessary
with the new way we're going to handle bundles.
-bin-bash_002


johnsca:
This looks like it should be fixed:

```
  File "/home/travis/build/conjure-up/conjure-up/conjureup/controllers/deploy/common.py", line 23, in do_deploy
    utils.spew(fn, app.current_bundle.to_yaml())
  File "/home/travis/build/conjure-up/conjure-up/conjureup/bundle.py", line 82, in to_yaml
    return yaml.dump(self._bundle,
AttributeError: 'Bundle' object has no attribute '_bundle'
```

PR #1389, opened 2018-04-03
Fix lxc query defaults

battlemidget:
The default is to perform a GET, which is all we use. The option changed from -X
to --X in the 2 to 3 switch.


tvansteenburgh:
LGTM

PR #1390, opened 2018-04-04
Fail with a error message on incompatible lxd versions

battlemidget:
Now only supporting LXD 3.x which can only be retrieved from snap store. Those
not using a snap now only have access to 2.0.x debs which conjure-up doesn't
support. Add a friendly error message and a link on how to install the required LXD.


battlemidget:
https://github.com/canonical-docs/conjure-up-docs/pull/34 for doc changes

PR #1391, opened 2018-04-04
Dont fail hard on incompatible LXD since it's an optional component

battlemidget:
This new way keeps conjure-up running and will perform the validation check in
the background.  While LXD is unavailable the user will still see the newly
defined message and be able to follow along with the steps outlined in the
online docs.

### DIFF
--- a/conjureup/ui/views/cloud.py
+++ b/conjureup/ui/views/cloud.py
@@ -17,12 +17,9 @@ class CloudView(BaseView):
     default_disabled_msg = 'This cloud is disabled due to your selection of ' \
                            'spell or add-on. Please use the arrow keys to ' \
                            'select another cloud.'
-    lxd_unavailable_msg = ("LXD not found, please install and wait "
-                           "for this message to disappear:\n\n"
-                           "  $ sudo snap install lxd\n"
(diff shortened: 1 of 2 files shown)


conjureup/ui/views/cloud.py
```diff
@@ -17,12 +17,9 @@ class CloudView(BaseView):
     default_disabled_msg = 'This cloud is disabled due to your selection of ' \
                            'spell or add-on. Please use the arrow keys to ' \
                            'select another cloud.'
-    lxd_unavailable_msg = ("LXD not found, please install and wait "
-                           "for this message to disappear:\n\n"
-                           "  $ sudo snap install lxd\n"
-                           "  $ sudo usermod -a -G lxd <youruser>\n"
-                           "  $ newgrp lxd\n"
-                           "  $ /snap/bin/lxd init")
+    lxd_unavailable_msg = ("conjure-up requires a newer version of LXD. "
```
tvansteenburgh: "LXD version 3.0.0 or greater is required. To install or upgrade, see https://docs.conjure-up.io/devel/en/#users-of-lxd"

PR #1444, opened 2018-05-24
Pass model-config options from conjurefile to juju

battlemidget:
Allows you to set things like `vpc-id` in conjurefile and be applied to your models.

Test with:

conjure-up -c test/Conjurefile
juju model-config and check extra-info item

### DIFF
--- a/conjureup/juju.py
+++ b/conjureup/juju.py
@@ -12,6 +12,7 @@ import yaml
 from juju.client.jujudata import FileJujuData
 from juju.controller import Controller
 from juju.model import Model
+from melddict import MeldDict
 
 from conjureup import consts, errors, events, utils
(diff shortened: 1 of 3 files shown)


conjureup/juju.py
```diff
@@ -239,6 +240,10 @@ def add_config(k, v):
         add_config("bootstrap-timeout", app.conjurefile['bootstrap-timeout'])
     if app.conjurefile['bootstrap-to']:
         cmd.extend(["--to", app.conjurefile['bootstrap-to']])
+    if app.conjurefile['model-config']:
```
battlemidget: This may not be needed as it would only apply during bootstrap to default models (im guessing just the controller model)

PR #1463, opened 2018-06-07
WIP: Adds support for bypassing juju and performing custom deploys

battlemidget:
Reuses the same addons/steps/deploy/status code with a few branches based on if
juju is required or not.

Could use some more refactoring and building on this, but this works and is a
good start i think.

Need spell branch: https://github.com/conjure-up/spells/tree/microk8s-spell


conjureup/models/step.py
```diff
@@ -281,7 +283,7 @@ def bundle_remove(self):
 
         # special case for 00_deploy-done to report masked
         # charm hook failures that were retried automatically
-        if not app.no_report:
+        if not app.no_report or app.metadata.needs_juju:
```
johnsca: Should this be `and` or `or not`?

PR #1466, opened 2018-06-08
File/Directory restructure for handling spell types

battlemidget:
Addition of spell-type in spell metadata allows us to alter the controller path depending on if a spell is juju or something else (currently snap). 

This PR lays the groundwork for expanding on those different controller paths and makes the metadata_controller setup spell-type specific as well (dealing with bundles vs other spell-type application definitions)


wwwtyro:
lgtm

PR #1467, opened 2018-06-08
Update controllers for spell-type snap

battlemidget:
This allows for new spell-types to be defined and to also outline what that controller journey would be.


battlemidget:
@johnsca this should be ready for review

conjureup/controllers/addons/gui.py
```diff
@@ -0,0 +1,38 @@
+from conjureup import controllers
```
johnsca: Maybe it would be better to put the shared controller logic under `conjureup/controllers/base` (or `common` if we want to match the existing convention, but I like `base` better, personally) so that it's more clear, and the `conjureup/controllers` directory only has `juju`, `snap`, and `common` (plus any future spell types we may add)?  I found the directory layout a bit confusing as I was reading through this.

PR #1471, opened 2018-06-15
Fix unable to select cloud when only public clouds are whitelisted

johnsca:
When selecting, e.g., CDK-NVIDIA with kubeflow and only AWS is white-listed, the cloud selection cursor is unavailable due to: 1) the list of clouds being mostly by name rather than type, while the white-list is by type, and 2) there being no fallback case for `select_item_by_value`.

Also fixes the isort issue by removing a unused widget code.


hyperbolic2346:
LGTM, thanks for the quick fix!

PR #1486, opened 2018-06-29
Python 3.7 compatibility

### LINKED ISSUE #1484: Python 3.7 support
We're working on the Python 3.7 PR for Homebrew and it looks like conjure-up is broken because `async` became a keyword but there's a conflicting conjureup.async module. There are also some issues with the dependencies (specifically juju needs a newer pyyaml and websockets).

CC @battlemidget 

It would be great if Python 3.7 support could be added ASAP here as this is blocking the 3.7 upgrade for everyone unless we're willing to break conjure-up.

battlemidget:
Fixes #1484


Makefile
```diff
@@ -71,20 +70,22 @@ clean:
...
-	tox -e conjure-dev
+	# NB: Remove once pyyaml 4.2 officially released
+	PIP_PRE=1 tox -e conjure-dev
 	@echo "Run 'source conjure-dev/bin/activate' to enter the dev venv"
 
 # Fix some of the python formatting preferred by pylint
 auto-format:
-	@tox -e isort -- isort -rc -m 3 conjureup test tools
-	@tox -e isort -- autopep8 --in-place --recursive conjureup test tools
+	PIP_PRE=1 tox -e isort -- isort -rc -m 3 conjureup test tools
```
battlemidget: going to remove this env var in next commit and just use PIP_PRE=1 make test for now

Makefile
```diff
@@ -7,22 +7,21 @@ TOPDIR := $(shell basename `pwd`)
 GIT_REV := $(shell git log --oneline -n1| cut -d" " -f1)
 VERSION := $(shell cat VERSION)
 CHANNEL := edge
-PY36 := $(shell apt-cache search --names-only '^python3.6')
+PY37 := $(shell apt-cache search --names-only '^python3.6')
```
johnsca: This line should end with `python3.7` no?

Makefile
```diff
@@ -71,20 +70,22 @@ clean:
 
 .PHONY: test
 test: auto-format
-	@tox -e py35,flake,isort
+	# NB: Remove once pyyaml 4.2 officially released
+	PIP_PRE=1 tox -e py35,flake,isort
```
johnsca: I think this should be `py37` as well.

requirements.txt
```diff
@@ -18,3 +18,4 @@ awscli==1.11.123
 pyvmomi==6.5.0.2017.5-1
 kv==0.3.0
 melddict==1.0.1
+git+https://github.com/aaugustin/websockets
```
johnsca: Comment to remove once 5.0.2 is released.

PR #1489, opened 2018-07-02
Add option for branch and force sync

johnsca:
The registry branch is supported being set via an env var, but not via the CLI or Conjurefile.  This adds support for it via both of those.

It also adds the ability to specify `--sync` via the CLI to override `--no-sync` previously being provided.  This is needed because the snap sets `--no-sync` via the CLI and this gives a way to override the snap.


hyperbolic2346:
LGTM, thanks for adding this.

PR #1560, opened 2018-12-13
Bump libjuju version to pick up subordinate fix

johnsca:
See juju/python-libjuju#277


johnsca:
Self-merging for release

PR #1581, opened 2019-02-04
Fix unable to select new controller

johnsca:


johnsca:
Tested this locally and it fixes the issue without breaking disabled controller handling.

PR #1655, opened 2021-04-05
Allow to define juju and juju-wait paths with env variables

iMichka:
This change allows to use juju and juju-wait when they are installed
somewhere else than in the standard paths.

For example on Linux, using the Homebrew package manager:
/home/linuxbrew/.linuxbrew/Cellar/juju

And on ARM macs:
/opt/Homebrew/Cellar

This allows to build and ship conjure-up with non-standard paths.

### DIFF
--- a/conjureup/juju.py
+++ b/conjureup/juju.py
@@ -44,10 +44,12 @@ def set_bin_path():
     """ Sets the juju binary path
     """
     candidates = [
+        os.environ['JUJU'],
         '/snap/bin/juju',
         '/snap/bin/conjure-up.juju',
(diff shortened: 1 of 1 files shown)


kwmonroe (approved):
LGTM, but please note that `conjure-up` is now end-of-life.  I'll merge this for anyone that may be maintaining a fork, but the main repo will be archived very soon.

Thank you!

iMichka:
Ok, thanks. No issue, I did not know it was soon EOL. We will deprecate the package then, and slowly phase it out. Thanks!.

